PR #1, opened 2021-11-19
prelim test of CI

cmharlow:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-128

What this PR does: 
* Adds tf-actions CI to this repository
* Adds the mozilla-it/terraform-modules gke module to here, primarily for testing CI & used for evolution as useful 

~This is blocked by https://github.com/mozilla/tf-actions/pull/1, which it uses for CI.~

PR #2, opened 2021-11-22
add first draft of gcp_project migrated module

cmharlow:
Jira: came up in context of https://mozilla-hub.atlassian.net/browse/OPST-166

What this PR does:
* Migrates https://github.com/mozilla-services/cloudops-infra-terraform-modules/tree/master/gcp-project to here for use in Global/Admin project work
* highlights some variables / metadata... we may no longer want? **including removing environment-specific metadata (repeated three times as environment, realm, and env_code) - since its unclear if we want environment metadata at the GCP Project level**
* updates required terraform version

Note:
* tf init & tf validate in module directory creates the .terraform.lock.hcl file... which i wasn't sure if we wanted to keep in modules specifically (it'll get ignored in usage in terraform projects as I understand it). [this Hashi discussion](https://discuss.hashicorp.com/t/what-is-the-best-practise-with-terraform-lock-hcl-in-modules-commit-or-not/28648) led me to checking it in though
* will update this PR with a sample of it being used shortly


cmharlow:
We might also want to rebase this PR once #1 is in so CI can check it too

google_project/variables.tf
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+
+variable "program_name" {
+  description = "Name of the Firefox program being one of: ci, data, infrastructure, services, web."
+  default     = "services"
+  type        = string
+
+  validation {
+    condition     = contains(["ci", "data", "infrastructure", "services", "web"], var.program_name)
```
kkleemola, line 82: Some of our projects use "security" as a program name.
(We can fix this later but might be good if @jasonthomas knows if there's any others we need)

kkleemola (dismissed):

google_project/variables.tf
cmharlow: yeah this is good point. just trying to use these validations as a forcing function for understanding what metadata we want going forward.

cmharlow, line 82:
```suggestion
    condition     = contains(["ci", "data", "infrastructure", "security", "services", "web"], var.program_name)
```

google_project/variables.tf
```diff
@@ -0,0 +1,85 @@
...
+  description = "Override default project id. Only use if the project id is already taken."
+  type        = string
+}
+
+variable "realm" {
+  description = "Project environment e.g., prod"
+  type        = string
+
+  validation {
+    condition     = contains(["dev", "global", "prod", "sandbox", "stage"], var.realm)
```
jasonthomas, line 56:
```suggestion
    condition     = contains(["global", "nonprod", "prod", "sandbox"], var.realm)
```
I don't think we want dev and stage realms, they would fall under nonprod in terms of projects, such that, all nonprod resources can be in one project.

google_project/variables.tf
jasonthomas: I think with the addition of security we should be good for now. I think we will eventually need to add others.

google_project/variables.tf
cmharlow: so, this isn't what we do in web SRE. is the idea that everyone would fall into prod / nonprod now? and does nonprod or prod include sandbox?

google_project/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  skip_delete = true
+
+  billing_account = var.billing_account_id
+  folder_id       = var.parent_id
+
+  auto_create_network = false
+
+  labels = {
+    app            = var.project_name
+    app_code       = local.app_code
```
jasonthomas, line 33: https://mana.mozilla.org/wiki/display/INEN/Firefox+Infrastructure+Inventory has more details around these different labels and how it is structured now.

google_project/variables.tf
jasonthomas: I'd like to understand the use cases/requirements around having separate projects for nonprod resources. Ideally we would just have two projects, nonprod and prod for each application/product.

I don't envision SRE using sandbox realm much, but it would have been useful for use cases when engineering teams use our module to provision 'sandbox' projects. https://github.com/mozilla-services/data-sandbox-terraform, which contains data engineering sandbox project resources, but instead, we were calling it something along the lines of 'dev' in the  project name and realm = nonprod.

jasonthomas: I looked through this again, and I realized that we are setting `environment = var.realm` for gcp project labels, which I don't think we should. We should set environment labels on the actual resources (if labels are supported), e.g. set environment label on dev and stage cloud sql instances when they are both in the nonprod project.

cmharlow: makes sense to me, will remove

google_project/main.tf
cmharlow: will review tomorrow and update this PR to reflect, thanks for sharing

google_project/variables.tf
cmharlow: added security

google_project/variables.tf
cmharlow: removed realm /environment, figuring it'll be setup on terraform projects specifically

cmharlow: removed all environment specific metadata from the project label.

google_project/main.tf
cmharlow: integrated into descriptions where it overlaps.

cmharlow:
Made this ticket for following up on metadata questions: https://mozilla-hub.atlassian.net/browse/OPST-179

kkleemola (approved):

google_project/variables.tf
jasonthomas: Whoops sorry, I was not clear here. We need to keep realm, but environment is good to remove.

PR #5, opened 2021-12-01
OPST-227: adding default google_project_service to google_project module

moz-astults:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-227

### DIFF
--- a/google_project/variables.tf
+++ b/google_project/variables.tf
@@ -73,3 +73,9 @@ variable "program_name" {
     error_message = "Valid values for program_name: ci, data, infrastructure, services, web."
   }
 }
+
+variable "project_services" {
+  description = "List of google_project_service APIs to enable."
+  default     = ["compute.googleapis.com", "container.googleapis.com", "dns.googleapis.com", "logging.googleapis.com", "monitoring.googleapis.com", "servicenetworking.googleapis.com", "stackdriver.googleapis.com", "iamcredentials.googleapis.com"]
+  type        = list(string)
+}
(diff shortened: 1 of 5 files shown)


google_project/variables.tf
```diff
@@ -73,3 +73,9 @@ variable "program_name" {
     error_message = "Valid values for program_name: ci, data, infrastructure, services, web."
   }
 }
+
+variable "project_services" {
+  description = "List of google_project_service APIs to enable."
+  default     = ["compute.googleapis.com", "container.googleapis.com", "dns.googleapis.com", "logging.googleapis.com", "monitoring.googleapis.com", "servicenetworking.googleapis.com", "stackdriver.googleapis.com", "iamcredentials.googleapis.com"]
```
jasonthomas, line 79: Having a set of defaults to enable on every project is a great idea, but I believe the default `project_services` should be moved into a local variable and this variable changed to `additional_project_services`, and the setunion of those two lists to enable the services so that it is easier to enable additional services and non have to copy-pasta all the services around again.

moz-astults: @jasonthomas thanks for this feedback! I'll work on this a bit later today and ping you when I have an update :slightly_smiling_face:

cmharlow: we have an example of that with the extra_project_labels fwiw. good point to not have the default leave us copy-pasting constantly.

moz-astults: @jasonthomas @cmharlow can you take another look at this and see if it makes sense?

jasonthomas (approved):

PR #6, opened 2021-12-09
google_project: add realm variable OPST-84

jasonthomas:
As part of the discussion https://github.com/mozilla/terraform-modules/pull/2#discussion_r755407063, I didn't clearly specify that we needed to keep the realm label. 

I believe only https://github.com/mozilla-it/global-platform-admin/blob/ac672e3eef0fd51054ddb991f1bab6f47d89b51e/shared/tf/projects.tf#L22-L32 is referring to this now and I will submit a PR for the change.

r?

### DIFF
--- a/google_project/README.md
+++ b/google_project/README.md
@@ -43,6 +43,7 @@ No modules.
 | <a name="input_project_id"></a> [project\_id](#input\_project\_id) | Override default project id. Only use if the project id is already taken. | `string` | `""` | no |
 | <a name="input_project_name"></a> [project\_name](#input\_project\_name) | Name of project e.g., autopush | `string` | n/a | yes |
 | <a name="input_project_services"></a> [project\_services](#input\_project\_services) | List of google\_project\_service APIs to enable. | `list(string)` | `[]` | no |
+| <a name="input_realm"></a> [realm](#input\_realm) | Realm is a grouping of environments being one of: global, nonprod, prod | `string` | `""` | no |
 
 ## Outputs
 
--- a/google_project/variables.tf
+++ b/google_project/variables.tf
@@ -79,3 +79,14 @@ variable "project_services" {
   default     = []
   type        = list(string)
 }
+
+variable "realm" {
+  description = "Realm is a grouping of environments being one of: global, nonprod, prod"
+  default     = ""
+  type        = string
+
+  validation {
+    condition     = contains(["global", "nonprod", "prod"], var.realm)
+    error_message = "Valid values for realm: global, nonprod, prod."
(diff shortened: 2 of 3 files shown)


google_project/README.md
```diff
@@ -5,48 +5,49 @@ Sets up a single GCP project linked to a billing account plus management metadat
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
```
jasonthomas, line 8: I am not exactly sure why the formatting changed here, I am using `terraform-docs version v0.16.0 1f686b1 linux/amd64` and pre-commit.

cmharlow: huh yeah, looks like it just made links

cmharlow: which is cool, whatevs, live & let live as long as there are docs i guess

google_project/variables.tf
```diff
@@ -78,4 +78,15 @@ variable "project_services" {
   description = "List of google_project_service APIs to enable."
   default     = []
   type        = list(string)
-}
\ No newline at end of file
+}
+
+variable "realm" {
+  description = "Realm is a grouping of environments being one of: global, nonprod, nonprod"
```
cmharlow, line 84:
```suggestion
  description = "Realm is a grouping of environments being one of: global, nonprod, prod"
```

cmharlow: what happens if a GCP project being created is not environment specific - that always defaults to global then?

cmharlow (approved):

moz-astults (approved):

google_project/variables.tf
jasonthomas: In what scenario were we would provision resources that aren't environment specific? I am assuming for something like shared resources? Then if so global should be used.

cmharlow: so we don't use a separate project per environment for web SRE atm. is the idea that this needs to switch then?

cmharlow: (still approving to move forward, but this is something that's been on my mind reviewing this terraform for cloudops-based mentalities)

cmharlow (approved):

google_project/variables.tf
jasonthomas: > so we don't use a separate project per environment for web SRE atm. is the idea that this needs to switch then?

Yes. From a security perspective, in the past (no record of this decision of-course :( ) we wanted segmentation between these environments. From a cost perspective, it made it easier to determine costs between them.

PR #10, opened 2022-01-27
initial checkin of cloudsql modules

cvalaas:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-261


cvalaas:
This definitely could be better. I updated the default DB versions for postgres and mysql. Also changed how machine types are specified.

google_cloudsql/postgres/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+}
+
+variable "enable_public_ip" {
+  default     = false
+  description = "If true, will assign a public IP to database instance."
+}
+
+variable "ip_configuration_require_ssl" {
+  default = false
+}
```
moz-astults, line 75: Any reason we shouldn't default this to `true`? Are there compatibility concerns we're worried about?

moz-astults (dismissed):
This is great! I appreciate the simple examples too

google_cloudsql/postgres/variables.tf
cvalaas: I think a lot of legacy code makes simple non-ssl connections. Which I also think is _probably_ okay in a private network scenario. A lot more okay than with a public endpoint, anyway.
That said, switching to default `true` is probably a smart idea.

cvalaas: Interesting - I thought this was for Mysql, but I see now it's for postgres. huh. I'll change the default and also look into mysql's config

moz-astults (approved):

bqbn:
Do you think we can have the two modules at the top level, instead of having both inside the `google_cloudsql` directory? It looks like other modules in this repo are at the top level anyways.

For MySQL module, some apps (amo for example) need more than one read replicas. Can we make the module be able to add more than one replica resources?

For the PostgreSQL module, is there a reason why it doesn't have a read replica resource in it?

google_cloudsql/mysql/main.tf
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  lifecycle {
+    prevent_destroy = true
+    ignore_changes  = [settings.0.disk_size]
+  }
+}
+
+resource "google_sql_database_instance" "replica" {
```
jbuck, line 84: This is the old way of doing HA - might be good to drop this and change to the new way (which I think is just setting a boolean attribute on the master instance)

Can read in depth at https://cloud.google.com/sql/docs/mysql/high-availability#legacy_mysql_high_availability_option

cvalaas:
Sorry, astults, I guess every push is going to wope out your approval :|

google_cloudsql/mysql/main.tf
cvalaas: Yeah, I just re-did that stuff a bit. HA is determined when you set `availability_type` to REGIONAL and replicas are a separate thing.

cvalaas:
As for postgres replicas, I'll have to look into it. Right now it's just a straight copy of what we had in our cloudops-terraform-modules

google_cloudsql/postgres/main.tf
```diff
@@ -0,0 +1,138 @@
...
+# https://github.com/terraform-providers/terraform-provider-google/issues/1773#issuecomment-408556558
+# for an example.
+
+locals {
+  default_database_name = "${var.application}-${var.realm}-${var.environment}-${var.instance_version}"
+  database_name         = coalesce(var.custom_database_name, local.default_database_name)
+  tier                  = coalesce(var.tier_override, "db-custom-${var.db_cpu}-${var.db_mem_gb * 1024}")
+}
+
+resource "google_sql_database_instance" "master" {
```
jasonthomas, line 21:
```suggestion
resource "google_sql_database_instance" "primary" {
```

google_cloudsql/mysql/main.tf
```diff
@@ -0,0 +1,130 @@
...
+
+  default_replica_name = "${local.database_name}-replica"
+  replica_name         = coalesce(var.custom_replica_name, local.default_replica_name)
+
+  ip_addresses = google_sql_database_instance.master.ip_address
+
+  enable_ha = var.realm == "prod" ? true : false
+}
+
+resource "google_sql_database_instance" "master" {
```
jasonthomas, line 19:
```suggestion
resource "google_sql_database_instance" "primary" {
```

google_cloudsql/mysql/main.tf
```diff
@@ -0,0 +1,130 @@
...
+        content {
+          expiration_time = lookup(authorized_networks.value, "expiration_time", null)
+          name            = lookup(authorized_networks.value, "name", null)
+          value           = lookup(authorized_networks.value, "value", null)
+        }
+      }
+    }
+
+    maintenance_window {
+      day          = "2"
```
jasonthomas, line 66: Note for future self, Tuesdays at 16:00 UTC.

google_cloudsql/postgres/main.tf
```diff
@@ -0,0 +1,138 @@
...
+  name     = local.database_name
+  project  = var.project_id
+  region   = var.region
+
+  database_version = var.database_version
+
+  settings {
+    tier              = local.tier
+    availability_type = var.availability_type
+    disk_size         = var.disk_size
```
jasonthomas, line 32: I believe `disk_autoresize` is optional but defaults true. In the documentation it says not to set disk_size since once auto resize kicks in terraform will try to resize the disk to disk_size value.

cvalaas: will fix

jasonthomas: So does this mean we just default to disk size (10GB) and let cloudsql handle the auto resize?

cvalaas: yeah, seems to be what terraform suggests

cmharlow:
> Do you think we can have the two modules at the top level, instead of having both inside the `google_cloudsql` directory? It looks like other modules in this repo are at the top level anyways.

Ditto to this. We're aiming for as flat as feasible, otherwise, this current nested setup would mess up versioning options down the road / post this migration MVP.

cvalaas:
> > Do you think we can have the two modules at the top level, instead of having both inside the `google_cloudsql` directory? It looks like other modules in this repo are at the top level anyways.
> 
> Ditto to this. We're aiming for as flat as feasible, otherwise, this current nested setup would mess up versioning options down the road / post this migration MVP.

Dir structure has been flattened.

google_cloudsql_mysql/main.tf
```diff
@@ -0,0 +1,126 @@
+/**
```
cmharlow, line 1: something missing here that is used by web sre terraform modules pretty heavily is setting a default application user for the database. is that worth adding in, or a reason to make those resources tf managed elsewhere?

google_cloudsql_mysql/main.tf
```diff
@@ -0,0 +1,126 @@
...
+  database_version = var.database_version
+
+  settings {
+    tier = local.tier
+
+    availability_type = var.availability_type
+
+    disk_type = "PD_SSD"
+    dynamic "database_flags" {
+      for_each = var.database_flags
```
cmharlow, line 31: clarifying for myself - this is where MySQL RDS parameter groups would be migrated to, right? Thinking of [param groups like used here](https://github.com/mozilla-it/itse-apps-prod-1-infra/pull/65/files), and it isn't clear to me in the [terraform docs for google_sql_database_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance) what flags are allowed or not.

google_cloudsql_mysql/main.tf
```diff
@@ -0,0 +1,126 @@
...
+
+  default_replica_name = "${local.database_name}-replica"
+  replica_name         = coalesce(var.custom_replica_name, local.default_replica_name)
+
+  ip_addresses = google_sql_database_instance.primary.ip_address
+
+  enable_ha = var.realm == "prod" ? true : false
+}
+
+resource "google_sql_database_instance" "primary" {
```
cmharlow, line 19: do we want to add any kind of ability to override `deletion_protection` for when we do want terraform to delete a db (i'd imagine it'd have to be a 2 PR process, one changing the deletion_protection default from true to false, a second to then ask atlantis to actually delete the instance) - or should we just leave deletion_protection outside of our terraform & defaulting to true, so databases have to be manually deleted?

google_cloudsql_postgres/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+  database_version = "POSTGRES_13"
+
+  availability_type = local.realm == "prod" ? "REGIONAL" : "ZONAL"
+
+  db_cpu    = 4
+  db_mem_gb = 20
+}
+
+output "postgres_database" {
```
cmharlow, line 100: same question as with mysql about creating application users also.

google_cloudsql_postgres/versions.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 0.13"
```
cmharlow, line 2: can this be 1.0 and above?

google_cloudsql_mysql/main.tf
```diff
@@ -0,0 +1,126 @@
...
+          name            = lookup(authorized_networks.value, "name", null)
+          value           = lookup(authorized_networks.value, "value", null)
+        }
+      }
+    }
+
+    maintenance_window {
+      day          = "2"
+      hour         = "16"
+      update_track = var.realm == "prod" ? "stable" : "canary"
```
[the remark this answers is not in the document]
cvalaas: moved

google_cloudsql_postgres/versions.tf
cvalaas: yes

google_cloudsql_postgres/versions.tf
```diff
@@ -0,0 +1,13 @@
...
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 3.58"
+    }
+    google-beta = {
```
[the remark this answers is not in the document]
cvalaas: don't think it does. copypasta. removing.

google_cloudsql_mysql/main.tf
cvalaas: added

google_cloudsql_mysql/main.tf
jbuck: Maybe we could add a link to the [supported flags](https://cloud.google.com/sql/docs/mysql/flags#list-flags-mysql) in the variable description?

google_cloudsql_mysql/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 3.0"
```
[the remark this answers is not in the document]
jbuck: Also would be good to do the same for the postgres module

cvalaas:
> Maybe we could add a link to the supported flags in the variable description?

Was there already. Updated link, tho. thanks!

google_cloudsql_mysql/main.tf
cvalaas: added

google_cloudsql_mysql/main.tf
cvalaas: do you have a link to an example of how websre did that? Just so i can copypasta instead of thinking.

cvalaas:
I vote we merge and work on user accounts as a new PR. Who's with me!?!

jbuck:
> I vote we merge and work on user accounts as a new PR. Who's with me!?!

I think that's reasonable - fwiw the reason we didn't tackle user account creation within Terraform and provisioned them manually is that the application db password would be stored in plaintext in the Terraform state. I don't know if that's still the case

google_cloudsql_mysql/main.tf
cmharlow: sure - see an example here: https://github.com/mozilla-it/terraform-modules/blob/0905ea5607511f9e57d867953030aa11db7e7989/gcp/database/main.tf#L28

google_cloudsql_mysql/main.tf
cmharlow: @jbuck you're certain the usage examples I pointed to are covered in those flags? that's basically what i'm asking, and your response doesn't seem clear if you just wanted to point to docs or actually knew the answer.

jbuck: @cmharlow yeah, here's `character_set_server` for example: https://cloud.google.com/sql/docs/mysql/flags#mysql-c (sorry, I can't link directly to it). The other variables aren't able to be set, which is why they're not listed there. I think the might inherit from `character_set_server` though?

google_cloudsql_mysql/main.tf
cmharlow: thank you!

google_cloudsql_mysql/main.tf
cmharlow: okay, so its unknown atm, but you're assuming it should just work somehow. will do some minor digging to confirm this.

jbuck: Confirmed you can't set unsupported (by CloudSQL) flags:

```
% gcloud sql instances patch --database-flags character_set_client=utf8mb4 --project moz-fx-balrog-nonprod-1841 balrog-nonprod-dev-v2
The following message will be used for the patch API method.
{"name": "balrog-nonprod-dev-v2", "project": "moz-fx-balrog-nonprod-1841", "settings": {"databaseFlags": [{"name": "character_set_client", "value": "utf8mb4"}]}}
WARNING: This patch modifies database flag values, which may require your instance to be restarted. Check the list of supported flags - https://cloud.google.com/sql/docs/mysql/flags - to see if your instance will be restarted when this patch is submitted.

Do you want to continue (Y/n)?  

ERROR: (gcloud.sql.instances.patch) HTTPError 404: The requested flag is either misspelled or unsupported by Cloud SQL.
```

cmharlow: alright, looks like its a known problem & hole in the GCP cloudsql docs. Basically, if we're on mysql 5.7 or above, setting `character_set_server` should set the parameters / type for the other columns, but its not consistent in its application & may require us to run some SQL to manually change the column types post database creation.

cmharlow (dismissed):
believe everything i've raised minus adding default db user / pword - which can be handled in a future PR per comments on that - has been been addressed

PR #11, opened 2022-01-31
move GCP app/project dns module to here

cmharlow:
Jira: came up in context of https://mozilla-hub.atlassian.net/browse/OPST-333

What this PR does:
* creates a new module for creating GCP Cloud DNS Zone & RecordSet when delegating a subdomain of some parent zone  in mozgcp.net (like testapp.nonprod.dataops.mozgcp.net, a subdomain of dataops.mozgcp.net) to an app-specific GCP Project's own Cloud DNS Zones
* basically replaces: https://github.com/mozilla-services/cloudops-infra-terraform-modules/tree/master/gcp-project-dns


google_project-dns/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+    error_message = "Valid values for realm: global, mgmt, nonprod, prod."
+  }
+}
+
+variable "team_name" {
+  description = "Name of SRE team, which should correspond to the top-level folder name"
+  type        = string
+
+  validation {
+    condition     = contains(["cloudops", "dataops", "websre"], var.team_name)
```
jasonthomas, line 37:
```suggestion
    condition     = contains(["servicessre", "dataops", "websre"], var.team_name)
```

google_project-dns/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+    error_message = "Valid values for team_name: cloudops, dataops, websre."
```
jasonthomas, line 38:
```suggestion
    error_message = "Valid values for team_name: servicessre, dataops, websre."
```

It's my understanding that we won't use cloudops, at least for now. We can also just have both cloudops and servicessre here as well.

jasonthomas (dismissed):

google_project-dns/variables.tf
cmharlow: yeah i'll add both to be safe, thanks!

cmharlow: @jasonthomas give it another review hwen you have a moment, thanks!

jasonthomas (approved):

PR #13, opened 2022-02-07
Add Redis terraform module

moz-astults:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-359


.github/workflows/ci.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
     - name: Get Changed Terraform directories
       id: search
       uses: mozilla/tf-actions/matrixify@main
+      with:
+        ignore_dir: "**/examples"
```
moz-astults, line 19: @cmharlow It seems like this works - you can look at my latest commits and see that once I added these lines linting passes. I then intentionally caused a formatting issue and the CI still fails as expected. If there's a better spot/convention for this ignore directory, let me know!

google_redis/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 3.0"
```
jbuck, line 5: Could you change this to 4.0?

google_redis/variables.tf
```diff
@@ -0,0 +1,68 @@
...
+variable "memory_size_gb" {
+  description = "Memory size in GiB"
+  default     = 1
+  type        = number
+}
+
+variable "redis_configs" {
+  description = "Redis configs https://cloud.google.com/memorystore/docs/redis/reference/rest/v1/projects.locations.instances#Instance.FIELDS.redis_configs"
+  type        = map(string)
+
```
jbuck, line 40: I wonder if we should add a default config for activedefrag=on?

google_redis/versions.tf
jbuck: Also, could you add the `google-beta` provider?

google_redis/variables.tf
```diff
@@ -0,0 +1,68 @@
...
+  description = "Environment e.g., stage."
+  type        = string
+}
+
+variable "realm" {
+  description = "Realm e.g., nonprod."
+  type        = string
+}
+
+variable "project_id" {
```
jbuck, line 27: I wonder if we should try to remove this variable, and switch to using the provider instead to determine what project it's installed in

google_redis/versions.tf
```diff
@@ -0,0 +1,9 @@
...
+    }
+  }
+  required_version = ">= 1.0.11"
```
jbuck, line 8: For Terraform versions is it better to use `>=` or `~>`?

google_redis/main.tf
```diff
@@ -0,0 +1,51 @@
...
+  connect_mode       = "PRIVATE_SERVICE_ACCESS" # Used for shared VPC access https://cloud.google.com/memorystore/docs/redis/networking
+
+  maintenance_policy {
+    weekly_maintenance_window {
+      day = var.maintenance_window_day
+      start_time {
+        hours   = var.maintenance_window_hour
+        minutes = 0
+        seconds = 0
+        nanos   = 0
```
jbuck, line 37: lol wow, I can't believe it's important to Google that you can schedule maintenance down to the nanosecond

jbuck (approved):
This looks good to me - I have some suggestions but only the version stuff really needs to be addressed. Really liking the terraform-docs customization that can be done too

google_redis/main.tf
moz-astults: lol yea this one was confusing to me as well

google_redis/variables.tf
moz-astults: @jbuck are there any performance considerations or any other reasons why someone would **not** want this enabled? Otherwise I can certainly rework this and add it as a default!

jbuck: I've had it enabled for FxA for years now and not run into any performance problems myself. I tried to look up anything about measuring the performance impact but only found "it's negligible" [here](https://serverfault.com/questions/971804/are-there-situations-were-activedefrag-should-be-kept-disabled-in-redis-5-with), and [here](https://stackoverflow.com/questions/61738210/redis-activedefrag-performance-reduction)

google_redis/variables.tf
moz-astults: good point, let me see if I can make it flexible enough for both

moz-astults: @jbuck I solved for this here: https://github.com/mozilla/terraform-modules/pull/13/commits/f4a515fb13ab2409488fd257d4b17d00776e2206

It looks like I can make the `project_id` variable's default value as `null` which will then use the provider block or env var etc. If it's not `null` it will use whatever was provided with that variable.

jbuck: Nice, that's elegant

google_redis/versions.tf
moz-astults: added!

google_redis/versions.tf
moz-astults: Good point, changed to `~>`

google_redis/variables.tf
moz-astults: Thanks! I added a `merge` and local variable so this is enabled by default. Since it is a `merge` people could override it by passing `{activedefrag: "no"}`

Added in this commit: https://github.com/mozilla/terraform-modules/pull/13/commits/7d3cf6ce102faff9662e7b0363b89825ffe5fb8c

jasonthomas (dismissed):

google_redis/README.md
```diff
@@ -0,0 +1,53 @@
+# Terraform Module: Redis
+Creates a Redis instance within GCP using Cloud Memorystore
+
+## Example
+
+```hcl
+locals {
+  name        = "test-redis"
+  realm       = "nonprod"
+  subnetworks = try(data.terraform_remote_state.vpc.outputs.subnetworks.realm[local.realm][local.project_id], {})
```
cmharlow, line 10: if an empty map is passed through, does a subnetwork get created?

cmharlow (approved):
one question, but looks like if an authorized network isn't provided, the module will just fail missing the required variable.

PR #18, opened 2022-02-09
gke wip

cmharlow:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-264

What this PR does:
* updates our GKE module to work with some defaults & as a private cluster within a Shared VPC (by default)
* TBD write up on the above and other changes


google_gke/variables.tf
```diff
@@ -1,101 +1,241 @@
...
+        ip_cidr_range = string
+        range_name    = string
+      })
+      service = object({
+        ip_cidr_range = string
+        range_name    = string
+      })
+    })
+    subnet_name = string
+    subnetwork  = string
```
jasonthomas, line 105: A heads up we are going to be exposing `subnetwork_id` since some gcp resources (like memcache) reference the network with the id. https://github.com/mozilla-it/global-platform-admin/pull/71

PR #22, opened 2022-02-17
Update GKE Module for Shared VPC, Private cluster, other assumptions

cmharlow:
Jira: https://mozilla-hub.atlassian.net/browse/OPST-264

What this PR does:
* Updates the GKE module (with Web SRE PoC updates so that this replaces #18 & informs [the forthcoming PRs](https://github.com/mozilla-it/websre-infra/pull/65) to replace https://github.com/mozilla-it/websre-infra/pull/63) with some assumptions and updates, discussed below.
* Adds some GKE examples that have _mostly_ been tested. Module has a lot of configurability though that has yet to be fully tested (see notes below).

Next steps:
* use this updated module in updates to various PoCs (~web~, data) & redeploy atlantis platform cluster to use this module & architectural changes.
* revisit setting up some default network policies to replace websre calico policies & leveraging (as feasible, this needs to be better evaluated as nobody in sre has used it yet) GKE dataplane v2 (which appears to still be calico under the hood).
* revisit list of GKE features to investigate further, noted in comment at top of cluster.tf file.

## About this work:

This module is reworked to default to the following:
* entirely private cluster (with option to opt out of this via variable overrides)
* cluster being run within a Shared VPC context (with option to opt out of this via variable overrides)
* the option to pass a Shared VPC output object to this module which can be parsed by the module to fill in all the networking variables
* require using a release channel (and not autopilot or no release channel)
* enabling & leverage (see next steps) gke dataplane v2 
* logging & monitoring configs setup to monitor control plane & workloads
* Daily maintenance time
* GCE Persistent Disk CSI Driver enabled
* Default max pods per node topping out at 32 (based on Shared VPC work)
* require using managed node pools instead of a cluster default node pool (which is removed upon cluster creation)
* option to export resource usage metrics to bigquery, with a big query dataset created specifically for this cluster if desired (this requires some further testing to confirm the connection, given gcp docs holes on this)
* ability to give cluster level perms to connect to other GCP Project Artifact Registries

This module removes:
* many of the configurability options available in previous GKE modules based on the above (e.g. autoplane, release channels, add-ons, etc.)
* all of the IT-SE / Web SRE bootstrapping (external-secrets, flux, velero, calico, configmapsecrets, etc.) until the group can decide on how this bootstrapping occurs
* single managed node pool at a time config from cloudops module (this module allows you to configure multiple node pools, so rollovers as done in web sre atm can be achieved, instead of a single managed node pool with a version suffix a la cloudops)

Some gotchas:
* testing of this has focused on using the module within a private cluster & shared vpc context - there are options to change both of these contexts, but they _have not been tested_ and i'm relying on them matching other terraform examples to indicate they're functional;
* this module leverages beta provider to get certain features we want (gce persistent disk csi driver; workloads level logging & monitoring configurability; linux sysctl node configurations) - HOWEVER, working with the google-beta provider did raise some bugs in testing (including one that required parsing a goroutine stacktrace and the provider go codebase). This is working now in testing, but we may want to discuss if these features are worth using beta.
* the parsing of the Shared VPC module output / data object for networking values requires this module be updated whenever that data shape changes (we're not using the experimental features of Terraform to allow for optional object keys in our variables, because this code will eventually be used in production).
* to have an entirely private cluster, we have to have at least one value in our master_authorized_networks_config; but I don't want to put Mozilla SRE private ips in this module as any defaults. So it defaults currently to a dummy value, but we may want to remove that default to force people to add an authorized networks config block; or, we add bastion ips to the networking outputs so it can be parsed from that.
* mentioned above, but I'm uncertain the bigquery export is functional, but the docs are not clear on if i need to set up any permissions to allow the export. I imagine a future PR will be required to repair or document the findings (@moz-astults see here).
* the node pools configurations (lists of maps) is a bit clunky when it comes to non-string configuration options (e.g. sysctls, labels, tags, taints, etc.) but it seems rare enough we use these (and we're hamstrung by terraform variables - see experimental features for objects aside above) that leaving as is


google_gke/cluster.tf
```diff
@@ -0,0 +1,206 @@
...
+      for_each = var.master_authorized_networks
+
+      content {
+        cidr_block   = lookup(cidr_blocks.value, "cidr_block", "")
+        display_name = lookup(cidr_blocks.value, "display_name", "")
+      }
+    }
+  }
+
+  # are we really using gke dataplane v2 (with limitations) over calico?
```
moz-astults, line 55: This piece is pretty confusing for me on how the terraform provider works, but I *think* we will just be using calico unless we actually set:
```
datapath_provider = "ADVANCED_DATAPATH"
```

Other than the scalability promised by dataplane v2, I'm not sure if we actually need it. However it looks like it's something that needs to be enabled at cluster creation so if we want it we should probably make the decision before standing up new clusters. The terraform provider docs are extremely lacking on this topic from what I've seen.

cmharlow: Are you confusing dataplane/calico and cilium / the data paths? Otherwise, not following what you're trying to understand here.

cmharlow: I changed the networking policy setup from Calico to this upon your motivation to use dataplane.

moz-astults: I definitely might be. It looks like to enable dataplane v2 you need to **not** specify the provider (like you did here) AND set the `datapath_provider`. I  haven't tested this yet though, so maybe that's a default now? Just looking at [this comment](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/656#issuecomment-910108440) and [this example](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest/examples/simple_regional_beta). 

When we had talked before the docs seem to indicate that dataplane v2 was "the new way" GKE was going to push people, but I wonder if that's not necessarily the case. Are there use-cases for calico that you think would make it worth using over dataplane v2? I'm honestly not that opinionated either way, it's just something that looks like it has to be set at cluster creation.

google_gke/variables.tf
```diff
@@ -1,101 +1,242 @@
...
+  type        = string
+
+  validation {
+    condition     = contains(["mgmt", "nonprod", "prod"], var.realm)
+    error_message = "Valid values for realm: mgmt, nonprod, prod."
+  }
+}
+
+variable "region" {
+  default     = "us-central1"
```
moz-astults, line 44: Do we want to set this to `null` so it will use the provider's default region?

README.md
```diff
@@ -91,3 +91,29 @@ Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary
...
+No resources.
+
+## Inputs
+
+No inputs.
+
+## Outputs
+
+No outputs.
+<!-- END_TF_DOCS -->
```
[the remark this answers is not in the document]
cmharlow: Nope, complete mistake on my part.

google_gke/cluster.tf
cmharlow: chatted with @moz-astults , have a list of questions to answer here. going to make the dataplane multi-field enablement a TF module variable so we can test.

google_gke/variables.tf
cmharlow: going to try shortly.

google_gke/locals.tf
```diff
@@ -1,69 +1,49 @@
...
+  project_id = trim(data.google_project.project.id, "/projects/")
+
+  tags_defaults = [var.realm, var.name, var.region, "terraform"]
+  tags          = setunion(local.tags_defaults, var.tags)
+
+  # internal networking setup
+  datapath_provider = var.enable_dataplane ? "ADVANCED_DATAPATH" : "DATAPATH_PROVIDER_UNSPECIFIED"
+
+  # monitoring setup
+  resource_usage_export_dataset_id = var.create_resource_usage_export_dataset ? google_bigquery_dataset.dataset[0].id : var.resource_usage_export_dataset_id
```
moz-astults, line 25: instead of `google_bigquery_dataset.dataset[0].id` it seems like this needs to be `google_bigquery_dataset.dataset[0].dataset_id`

From my testing, `id` gives the whole path(`projects/{{project}}/datasets/{{dataset_id}}`) but the module seems to only want the `dataset_id` and won't accept the whole path. Confirmed my testing works with the suggestion above.

google_gke/cluster.tf
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  upgrade_settings {
+    max_surge       = each.value.max_surge
+    max_unavailable = each.value.max_unavailable
+  }
+
+  lifecycle {
+    create_before_destroy = true
+  }
```
moz-astults, line 211: In my testing using an example like `complex1`, I needed to add the following, otherwise the node pool would try to be re-created:
```
    ignore_changes = [
      node_config[0].oauth_scopes,
      node_config[0].metadata,
    ]
```

cmharlow: cool thanks will add.

google_gke/examples/complex1.tf
```diff
@@ -0,0 +1,37 @@
...
+  create_resource_usage_export_dataset = true
+
+  # access docker image GARs in another project
+  # (self-same cluster project id included by default)
+  registry_project_ids = [
+    "team-app1"
+  ]
+
+  # who can access the k8s control plane
+  # adds currently known, internal bastions by default
```
moz-astults, line 29:
```suggestion
  # adds placeholder bastion network by default
```

google_gke/examples/complex2.tf
```diff
@@ -0,0 +1,50 @@
...
+  ]
+
+  node_pools_labels = {
+    nodepool-2 = {
+      "my-np2-label" = "some-value"
+    }
+  }
+
+  # who can access the k8s control plane
+  # adds currently known, internal bastions by default
```
moz-astults, line 43:
```suggestion
  # adds placeholder bastion network by default
```

google_gke/examples/simple.tf
```diff
@@ -0,0 +1,35 @@
...
+  network                     = "projects/my-vpc-project/global/networks/my-vpc-network"
+  pods_ip_cidr_range_name     = "my-pods-or-cluster-secondary-range-name"
+  services_ip_cidr_range_name = "my-services-secondary-range-name"
+  subnetwork                  = "projects/my-vpc-project/regions/us-west1/subnetworks/my-subnetwork"
+
+  # don't expect metrics to BQ
+  enable_resource_consumption_export = false
+
+  # who can access the k8s control plane
+  # adds currently known, internal bastions by default
```
moz-astults, line 28:
```suggestion
  # adds placeholder bastion network by default
```

google_gke/cluster.tf
cmharlow: this is a variable now, will work as a follow-up (which this was always listed as for this PR) to confirm what astults wants and what is needed to migrating existing network policy usage.

google_gke/cluster.tf
```diff
@@ -0,0 +1,212 @@
...
+  monitoring_config {
+    enable_components = [
+      "SYSTEM_COMPONENTS",
+      "WORKLOADS"
+    ]
+  }
+
+  dynamic "resource_usage_export_config" {
+    for_each = var.resource_usage_export_dataset_id != null ? [{
```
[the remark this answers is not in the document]
cmharlow: changed.

google_gke/locals.tf
cmharlow: changed

moz-astults (approved):

PR #26, opened 2022-03-14
google_gke: using project_id directly and requiring a value for var.project_id

moz-astults:


google_gke/variables.tf
```diff
@@ -25,7 +25,6 @@ variable "name" {
 }
 
 variable "project_id" {
-  default     = null
```
moz-astults, line 28: If we still want to be able to use the provider's `project_id` (accomplished with the default being `null`) we would need to go through the other areas of this module to replace any `local.project_id` references interpolated as part of a string and use something else. For now, I propose we just require this to be set.

moz-astults:
Thinking about this more, I actually don't think we need the `local.project_id` variable at all if we're going to require `var.project_id` to be set. I added another commit to replace the instances of the local variables.

cmharlow (approved):
lgtm after discussion of why is this occurring in the first place

moz-astults:
Closing this since we found the root cause - terraform's `trim` function is tricky...

PR #27, opened 2022-03-15
Using the project_id variable directly rather than trimming the id

moz-astults:
A better fix replacing https://github.com/mozilla/terraform-modules/pull/26


cmharlow (approved):
thank you for doing this

PR #32, opened 2022-06-07
terraform: Remove provider lock files

jbuck:


smarnach (approved):

smarnach:
I agree that this lock file is not used by users of the modules, but I still think it's better not to include the lock files here. It's more useful to let the test reinitialize Terraform with the latest version whenever they run, because this best simulates what happens in new usages of the modules. Otherwise, we won't learn when upstream updates break our modules.

smarnach:
The general rule is that the lock file should always be included in root modules, but not in other modules.

jbuck:
> @jbuck that's not true, its only used for testing the module alone. I can find the link that guided us on including these when we originally did the terraform work / modules setup if useful.

I believe the problem that we were running into when working on a separate PR (which I should've linked here, because now I'm not sure *which* PR I was having problems) was that it would always install the v3.x version of the Google provider when the separate PR needed to use the v4.x version. I think this PR is what fixed the problem, but I could be completely wrong here and this needs to be reverted

smarnach:
I believe the PR that fixed our problem was #33, not this one. I still believe this one shouldn't be reverted. Is this causing any problems for you, @cmharlow?

PR #41, opened 2022-07-14
adding gsm module

cvalaas:
This is the GKE service account stuff. other bootstrapping will be in a module which calls this one


google_gke_tenant/gke_service_account.tf
```diff
@@ -0,0 +1,33 @@
...
+  members = [
+    "serviceAccount:${var.cluster_project_id}.svc.id.goog[${var.namespace}/external-secrets]"
+  ]
+}
+
+# permissions for use with External Secrets Operator in GKE
+#
+resource "google_project_iam_member" "sa-role-token-creator" {
+  project = var.project_id
+  role    = "roles/iam.serviceAccountTokenCreator"
```
jbuck, line 19: Is this role required? In [the docs](https://external-secrets.io/v0.5.8/provider-google-secrets-manager/#using-pod-based-workload-identity) it says:

> You need to give the Google service account the `roles/iam.serviceAccountTokenCreator` role so it can generate a service account token for you (not necessary in the Pod-based Workload Identity bellow)

If you wanted to keep this role for testing while you set it up, could you add a `FIXME` type comment so we know to come back to this?

google_gke_tenant/gke_service_account.tf
```diff
@@ -0,0 +1,33 @@
...
+  member  = "serviceAccount:${google_service_account.gke-account.name}"
+}
+
+resource "google_project_iam_member" "sa-role-secret-viewer" {
+  project = var.project_id
+  role    = "roles/secretmanager.viewer"
```
jbuck, line 25: Same q as above, [the docs](https://external-secrets.io/v0.5.8/provider-google-secrets-manager/#creating-workload-identity-service-accounts) say:

> ROLE_NAME: should be `roles/secretmanager.secretAccessor` - so you make the pod only be able to access secrets on Secret Manager

jbuck (approved):
r+ with nits

google_gke_tenant/gke_service_account.tf
cvalaas: TBH, I didn't really understand the "Using Pod-based Workload Identity" section, so I went with the "Using Service Accounts directly" instructions above it.
Reading it again, it seems like maybe that'd require a container for ESO in every pod? Or only one service account for all namespaces? I don't get it. :(

google_gke_tenant/gke_service_account.tf
cvalaas: Ah yes, I think I had to add this to make the tag-based secrets ingestion work (since it needs to list all the secrets), but since we're not doing that now, I can remove.

google_gke_tenant/gke_service_account.tf
cvalaas: added FIXME

jasonthomas (approved):

PR #47, opened 2022-07-29
google_workgroup: add workgroup module OPST-682

jasonthomas:
Jira Ticket: https://mozilla-hub.atlassian.net/browse/OPST-682

r?

### DIFF
--- a/google_workgroup/main.tf
+++ b/google_workgroup/main.tf
@@ -1,0 +1,93 @@
+/**
+ * # workgroup
+ * Retrieve workgroup ACL lists associated with data and gcp access workgroups.
+ *
+ * Workgroup identifiers should be of the form:
+ *
+ * ```
+ * workgroup:WORKGROUP_NAME[/SUBGROUP]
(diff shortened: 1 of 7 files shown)


google_workgroup/main.tf
```diff
@@ -0,0 +1,93 @@
...
+ * subgroup:SUBGROUP
+ * ```
+ *
+ * is supported, which will return all workgroups that contain a particular subgroup.
+ *
+ * This module is cloned from https://github.com/mozilla-services/cloudops-infra-terraform-modules/tree/master/data-workgroup.
+ *
+ */
+
+data "terraform_remote_state" "workgroups" {
```
whd, line 25: This is simpler than https://github.com/mozilla-services/cloudops-infra-terraform-modules/commit/ff6c3035468fc003d58ecee5c14247401240985b, I guess I did it this way before because the primary use case was local testing (no remote state)

whd (approved):

PR #53, opened 2022-10-28
aws_gcp_vpn: Initial commit

jbuck:


aws_gcp_vpn/main.tf
```diff
@@ -0,0 +1,163 @@
...
+  bgp_asn    = var.gcp_private_asn
+  ip_address = each.value.ip_address
+  type       = "ipsec.1"
+
+  tags = {
+    Name = "gcp-${var.gcp_project_id}-to-aws-cg${each.value.id}"
+  }
+}
+
+# FIXME This only needs to be created once per AWS account
```
jasonthomas, line 78: If it's once per account then maybe outside of this module and add it as a documentation requirement? We just need to aws_vpn_gateway.id as an input.

jbuck: mmm yeah

jasonthomas: Or maybe just a resource that can be flagged on or off.

jasonthomas (dismissed):
r+wc

jasonthomas (approved):

PR #55, opened 2022-11-03
Allow the redis module to use newer terraform version

bqbn:


bqbn:
I've tested it working with terraform v1.3.3.

jbuck (approved):

PR #56, opened 2022-11-04
feat: copy google-cdn-external from cloudops-infra-terraform-modules

mikaeld:
# Description
This PR copies `google-cdn-external` from [cloudops-infra-terraform-modules](https://github.com/mozilla-services/cloudops-infra-terraform-modules). It will be used in [webservices-infra/data-static-websites](https://github.com/mozilla-it/webservices-infra/tree/main/data-static-websites)

I was thinking about linking this modules as a git submodule in [cloudops-infra-terraform-modules](https://github.com/mozilla-it/webservices-infra/tree/main/data-static-websites) so we don't have to maintain both modules. Looking for feedback on this

### DIFF
--- a/google_cdn-external/README.md
+++ b/google_cdn-external/README.md
@@ -1,0 +1,56 @@
+# Google CDN Distribution for external endpoints
+
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.13 |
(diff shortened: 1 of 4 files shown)


jasonthomas (approved):

jasonthomas:
@mikaeld it looks like the commit needs to be signed, can you do that?

jasonthomas (approved):

mikaeld:
Alright, time to configure GPG signing in my global git config instead of per project 😅

PR #61, opened 2022-11-10
Added risk_level as a label variable

mlarsonweber:
Updated the locals and variables files to add risk_level as a label

### DIFF
--- a/google_project/variables.tf
+++ b/google_project/variables.tf
@@ -48,6 +48,12 @@ variable "extra_project_labels" {
   default     = {}
 }
 
+variable "risk_level" {
+  default     = "low"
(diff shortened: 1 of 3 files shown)


google_project/variables.tf
```diff
@@ -48,6 +48,12 @@ variable "extra_project_labels" {
   default     = {}
 }
 
+variable "risk_level" {
+  default     = "low"
```
jasonthomas, line 52:
```suggestion
  default     = ""
```
I think this should be an empty string (to allow for backward compatibility) or no default at all.

jasonthomas (approved):

PR #69, opened 2022-12-12
google_gke: add monitoring_config_enable_components variable with def…

jasonthomas:
…ault values

Adds support to select which monitoring components to enable on a cluster basis. This is in prepartion for deprecation of `WORKLOADS` in GKE > 1.24.0
```
Error: googleapi: Error 400: Workload monitoring configuration is not supported for cluster version 1.24.5-gke.600; a version above 1.20.0 and below 1.24.0 is required., badRequest
```

[1] https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#enable_components


dlactin (approved):
lgtm!

PR #80, opened 2023-01-30
feat(google_cloudsql_mysql): add deletion_protection_enabled flag

ahoneiser:


cvalaas (approved):

google_cloudsql_mysql/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = "~> 4.0"
```
ahoneiser, line 5: @cvalaas updated the version constraint today, change makes sense for >= 4.48.0 only

cvalaas (approved):

PR #83, opened 2023-02-08
feat(google_datastream): Adding a stupid google datastream module

cvalaas:


google_datastream/main.tf
```diff
@@ -0,0 +1,72 @@
...
+    realm          = var.realm
+  }
+
+  vpc_peering_config {
+    vpc    = var.vpc_network
+    subnet = var.datastream_subnet
+  }
+}
+
+#resource "google_datastream_connection_profile" "source_connection_profile" {
```
jasonthomas, line 23: How does someone make use of this since it is commented out? Maybe this could put as part of the examples instead?

cvalaas: It can't be used unless we want to put the actual db password in terraform. The creation of the source profile and the stream itself have to be done manually :(

jasonthomas: Okay makes sense. Could we add a comment or details around that in the README or next to the resource definition?

jasonthomas (approved):
r+wc 

LGTM if you could just add some notes around the resources that need to be manually manipulated that would be great.

jasonthomas (approved):

PR #86, opened 2023-02-16
feat(google_project): Data access logging

mlarsonweber:
Added a resource to enable data access logs if the project has a risk level label of high

### DIFF
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -32,7 +32,7 @@ resource "google_project_service" "project" {
 }
 
 resource "google_project_iam_audit_config" "data_access_high" {
-  count = local.labels.risk_level == "high" ? 1 : 0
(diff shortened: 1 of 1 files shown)


jasonthomas:
We switched to [semantic versioning](https://github.com/mozilla/terraform-modules#versioning-and-releases), can you either update your commit or PR title to match?

mlarsonweber:
I haven't used that before, it's neat. Check is passing now!

jasonthomas (approved):

jasonthomas:
I realized that we missed excluding the data access logs from the project, we want to do this so we don't incur the cost and these are getting routed later in the org logging sink. 

We'd need a resource like https://github.com/mozilla-it/global-platform-admin/blob/main/modules/gke_log_router/main.tf#L23-L29 - but excluding all data access logs except for bigquery and google secrets manager, and IAM.

PR #87, opened 2023-02-17
feat(google_project): Data access logging exclusion

mlarsonweber:
Let me know how this looks, I wasn't sure what the @type should be for secretmanager or iam so I tried to use serviceName which I believe should work. I didn't make this conditional on the risk level, so every project would get this exclusion, regardless of whether the logs exist or not.

### DIFF
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -49,8 +49,8 @@ resource "google_project_iam_audit_config" "data_access_high" {
 
 resource "google_logging_project_exclusion" "data_access_exclusions" {
 
-  name        = "exclude-data-access-${each.key}"
-  description = "Exclude data access logs except BigQuery, secrets manager, and IAM for ${each.key}"
+  name        = "exclude-data-access-log-sink"
+  description = "Exclude data access logs except BigQuery, secrets manager, and IAM for this project"
 
   filter = <<EOT
 cloudaudit.googleapis.com/data_access
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -32,7 +32,7 @@ resource "google_project_service" "project" {
 }
 
 resource "google_project_iam_audit_config" "data_access_high" {
-  count = local.labels.risk_level == "high" ? 1 : 0
+  count = var.labels.risk_level == "high" ? 1 : 0
 
   project = local.project_id
   service = "allServices"
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -32,7 +32,7 @@ resource "google_project_service" "project" {
 }
 
 resource "google_project_iam_audit_config" "data_access_high" {
-  count = var.labels.risk_level == "high" ? 1 : 0
+  count = var.risk_level == "high" ? 1 : 0
 
   project = local.project_id
   service = "allServices"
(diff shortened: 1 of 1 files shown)


google_project/main.tf
```diff
@@ -29,4 +29,34 @@ resource "google_project_service" "project" {
...
+  audit_log_config {
+    log_type = "ADMIN_READ"
+  }
+  audit_log_config {
+    log_type = "DATA_READ"
+  }
+  audit_log_config {
+    log_type = "DATA_WRITE"
+  }
+}
```
jasonthomas, line 48: This resource is already available on main, we just need to update the branch to include this.

google_project/main.tf
```diff
@@ -29,4 +29,34 @@ resource "google_project_service" "project" {
...
+
+resource "google_logging_project_exclusion" "data_access_exclusions" {
+
```
jasonthomas, line 51: nit: additional whitespace

google_project/main.tf
```diff
@@ -29,4 +29,34 @@ resource "google_project_service" "project" {
...
+  name        = "exclude-data-access-log-sink"
+  description = "Exclude data access logs except BigQuery, secrets manager, and IAM for this project"
+
+  filter = <<EOT
+cloudaudit.googleapis.com/data_access
+AND NOT protoPayload.metadata."@type"="type.googleapis.com/google.cloud.audit.BigQueryAuditMetadata"
+AND NOT protopayload.metadata."serviceName"="secretmanager.googleapis.com"
```
jasonthomas, line 58:
```suggestion
AND NOT protoPayload.serviceName="secretmanager.googleapis.com"
```
The fields are case sensitive, and serviceName is available as a field in protoPayload

google_project/main.tf
```diff
@@ -29,4 +29,34 @@ resource "google_project_service" "project" {
...
+AND NOT protopayload.metadata."serviceName"="iam.googleapis.com"
```
jasonthomas, line 59:
```suggestion
AND NOT protopayload.metadata."serviceName"="iam.googleapis.com"
```
AND NOT protoPayload.serviceName="iam.googleapis.com"

google_project/main.tf
```diff
@@ -29,4 +29,34 @@ resource "google_project_service" "project" {
...
+  }
+}
+
+resource "google_logging_project_exclusion" "data_access_exclusions" {
+
+  name        = "exclude-data-access-log-sink"
+  description = "Exclude data access logs except BigQuery, secrets manager, and IAM for this project"
+
+  filter = <<EOT
+cloudaudit.googleapis.com/data_access
```
jasonthomas, line 56:
```suggestion
log_id("cloudaudit.googleapis.com/data_access")
```
Should we wrap this with log_id?

jasonthomas:
> Let me know how this looks, I wasn't sure what the @type should be for secretmanager or iam so I tried to use serviceName which I believe should work. I didn't make this conditional on the risk level, so every project would get this exclusion, regardless of whether the logs exist or not.

I made some comments, serviceName should work. I don't think we need to make it conditional on risk level here. We can always extend as needed.

jasonthomas (approved):

PR #94, opened 2023-03-02
feat: CDN Updates

bkochendorfer:
This adds some CDN configuration options that I will need to implement the Bedrock CDN using this shared module. I also added in this document yaml to regenerate. Hopefully that is the right thing or I can back out adding that.

### DIFF
--- a/google_cdn-external/main.tf
+++ b/google_cdn-external/main.tf
@@ -30,9 +30,9 @@ resource "google_compute_backend_service" "default" {
 
   protocol = var.origin_protocol
 
-  custom_request_headers = [
-    "host: ${var.origin_fqdn}"
-  ]
+  custom_request_headers = concat(["host: ${var.origin_fqdn}"], var.custom_request_headers)
+
+  custom_response_headers = var.custom_response_headers
 
   backend {
     group = google_compute_global_network_endpoint_group.default.self_link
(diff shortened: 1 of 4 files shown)


google_cdn-external/variables.tf
```diff
@@ -115,3 +115,15 @@ variable "quic_override" {
...
+variable "custom_response_headers" {
+  description = "Specifies custom response headers"
+  type        = list(string)
+  default     = null
+}
+
+variable "custom_request_headers" {
+  description = "Specifies custom response headers"
+  type        = list(string)
+  default     = null
```
jasonthomas, line 128:
```suggestion
  default     = []
```
concat doesn't seem to evaluate null properly
```
> concat(["test"], null)

> ╷
│ Error: Invalid function argument
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Invalid value for "seqs" parameter: argument must not be null.
╵

> concat(["test"], [])
[
  "test",
]
```

google_cdn-external/variables.tf
```diff
@@ -115,3 +115,15 @@ variable "quic_override" {
   type        = string
   default     = "DISABLE"
 }
+
+variable "custom_response_headers" {
+  description = "Specifies custom response headers"
+  type        = list(string)
+  default     = null
```
jasonthomas, line 122:
```suggestion
  default     = []
```
Probably can also change it here too.

google_cdn-external/variables.tf
bkochendorfer: That is a great catch @jasonthomas

jasonthomas (approved):

google_cdn-external/main.tf
```diff
@@ -55,9 +55,13 @@ resource "google_compute_backend_service" "default" {
       serve_while_stale            = lookup(var.cdn_policy, "serve_while_stale", null)
       signed_url_cache_max_age_sec = lookup(var.cdn_policy, "signed_url_cache_max_age_sec", null)
       cache_key_policy {
-        include_host         = lookup(var.cache_key_policy, "include_host", true)
-        include_protocol     = lookup(var.cache_key_policy, "include_protocol", true)
-        include_query_string = lookup(var.cache_key_policy, "include_query_string", true)
+        include_host           = lookup(var.cache_key_policy, "include_host", true)
+        include_protocol       = lookup(var.cache_key_policy, "include_protocol", true)
+        include_query_string   = lookup(var.cache_key_policy, "include_query_string", true)
+        include_http_headers   = lookup(var.cache_key_policy, "include_https_heaers", null)
```
cvalaas, line 61: typo on this line?

cvalaas (changes requested):

google_cdn-external/main.tf
cvalaas: "heaers"

cvalaas: also the var is include_HTTP_headers while the lookup is for include_HTTPS_...

cvalaas: (maybe that last one is intentional?)

bkochendorfer: Just a typo! Thanks for catching

cvalaas (dismissed):

bkochendorfer:
I added a couple more things here to get this to work with cert-manager. I'm still waiting to see if I can move forward with this option for my project in GCP.

jasonthomas (approved):

bkochendorfer:
@jasonthomas I ended up not being able to use this module and maybe also not use the CDN haha. I think I'll close this for now, we can always resurrect if needed.

PR #99, opened 2023-03-16
fix(google_project): do not exclude sts logs from project logging

jasonthomas:
r?


mlarsonweber:
This is to exclude the STS logs from data access?

jasonthomas:
The wording is confusion, this is to add back STS logs so that they are available in gcp project logging. This doesn't change anything related to the data access org sink.

mlarsonweber:
Ah, that makes sense, lgtm

mlarsonweber (approved):

PR #108, opened 2023-05-03
feat(cdn-external): support externally specified bucket backends

whd:
This PR is uglier than it looks, but it maintains API compatibility with the existing module while allowing additional path rules to be specified for a single `host_rule`. In this case, we want to divert some paths to a (not-external) storage backend while having the default still be the external backend. There's certainly an argument that this feature doesn't belong in this module but then I would need to redo `analysis-output.tmo` to use a custom module again which seems like it would be a lot more code overall.

I just picked the version of terraform I have on my local machine to get CI to pass. I should set it to the lowest value that `optional` became part of the main language, which might have been `1.3.0` but I will double check. I suspect the PR I'm about to file that uses this is going to have the same problem.


whd:
I think this is ready for review. I've tagged people who has worked on this module for feedback, and I'm very open to changing the way this works if anyone can come up with something less ugly that is API-compatible.

I've tested this module in all `backend_type` modes and it _should_ work but I would not be surprised if my test resources don't match existing deployments for some reason.

google_cdn-external/main.tf
```diff
@@ -3,21 +3,26 @@
...
 
 locals {
-  name_prefix             = join("-", [var.application, var.environment, var.name != "" ? "${var.name}-cdn" : "cdn"])
-  url_map_default_service = var.backend_type == "bucket" ? try(google_compute_backend_bucket.default[0].id, "") : try(google_compute_backend_service.default[0].id, "")
-  url_map_self_link       = var.backend_type == "bucket" ? try(google_compute_backend_bucket.default[0].self_link, "") : try(google_compute_backend_service.default[0].self_link, "")
+  name_prefix = join("-", [var.application, var.environment, var.name != "" ? "${var.name}-cdn" : "cdn"])
+  # when both a bucket and backend service are specified, prefer the backend
+  # service as the default backend, and use backend_bucket_paths to route
+  # specific paths to the backend bucket
+  url_map_default_service        = var.backend_type == "bucket" ? one(google_compute_backend_bucket.default[*].id) : one(google_compute_backend_service.default[*].id)
```
emaydeck-mozilla, line 10: Nice, this is definitely a cleaner approach.

google_cdn-external/main.tf
```diff
@@ -3,21 +3,26 @@
...
+  url_map_self_link              = var.backend_type == "bucket" ? one(google_compute_backend_bucket.default[*].self_link) : one(google_compute_backend_service.default[*].self_link)
+  backend_bucket_default_service = one(google_compute_backend_bucket.default[*].id)
```
emaydeck-mozilla, line 12: I'm not seeing this referenced anywhere. Do you expect you'll need to use it at some point?

whd: Good catch. I will double check but I think I might have meant to reference it in the case where both a service and backend bucket were specified, in which case I'll update it to be referenced. Otherwise I'll remove it.

emaydeck-mozilla (approved):

google_cdn-external/main.tf
whd: Removed. It's a tad confusing that the `default_service` parameter at https://github.com/mozilla/terraform-modules/blob/main/google_cdn-external/main.tf#L121 and the one at https://github.com/mozilla/terraform-modules/blob/main/google_cdn-external/main.tf#L135 take different representations, but that's a Terraform/GCP thing. In any case for bucket `backend_bucket_paths` only the `self_link` version is needed.

emaydeck-mozilla: Gotcha, makes sense. Approved the PR.

whd: whoops looks like I re-requested review after you had already approved, can you re-approve?

emaydeck-mozilla (approved):

google_cdn-external/main.tf
emaydeck-mozilla: np, done

PR #112, opened 2023-05-26
feat(google_cloudsql_postgres): Allow control of enable_private_path_…

bkochendorfer:
…for_google_cloud_services


jbuck (approved):
lgtm! I wonder if it'd be safe to set the default to true though?

bkochendorfer:
My gut here is to disable unless someone intentionally wants to expose that. This should be mostly fine but erring on the side of less access.

PR #114, opened 2023-06-27
feat(google_project): import _Default Logs Storage Bucket config

ahoneiser:


ahoneiser:
:information_source: use [google_logging_project_bucket_config](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/logging_project_bucket_config) resource to reduce the retention days for logs .. I'd like to make use of this to reduce the log storage for projects like remote-settings that hold a ton of logs e.g. for CDNs without us ever looking at more than a week worth of logs in case of errors ..

ahoneiser:
looks like storing logs <= 30 days (the default) is included in the ingestion fee .. so no need to care for lower log retention periods from a cost perspective :+1:

PR #117, opened 2023-07-06
feat: allow deployments from multiple Github environments

bqbn:
The default deployment service accounts, i.e. deploy-{dev,stage,prod}, are each tied to a specific environment. Therefore, they can only deploy from the respective tied environment. However, there are instances where we need to deploy from multiple environments using the same service account. This pull request aims to address this specific use case.

### DIFF
--- a/google_deployment_accounts/main.tf
+++ b/google_deployment_accounts/main.tf
@@ -12,7 +12,9 @@ resource "google_service_account" "account" {
 resource "google_service_account_iam_binding" "github-actions-access" {
   service_account_id = google_service_account.account.name
   role               = "roles/iam.workloadIdentityUser"
-  members = [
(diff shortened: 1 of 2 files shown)


hwine:
If this is a global change, it needs an RRA. See [slack thread](https://mozilla.slack.com/archives/C023TE3GE65/p1688659403845009)

bqbn:
@jasonthomas @hwine please see the comment I added to this PR. The language is general since this is a public repo. I plan to add more doc to the PR that I will file for the [tenant_bootstrap](https://github.com/mozilla-it/webservices-infra/tree/main/modules/tenant_bootstrap) module with the statements that Hal proposed.

But let me know if the wording is OK in this PR. Thanks!

jasonthomas (approved):

PR #138, opened 2023-10-18
feat: Expose log_analytics variable and create _Default bucket in TF …

Flipez:
…if true

### DIFF
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -60,3 +60,11 @@ AND NOT protoPayload.serviceName="iam.googleapis.com"
 AND NOT protoPayload.serviceName="sts.googleapis.com"
   EOT
 }
+
+resource "google_logging_project_bucket_config" "project" {
+  count            = var.log_analytics ? 1 : 0
+  project          = local.project_id
+  location         = "global"
+  bucket_id        = "_Default"
+  enable_analytics = var.log_analytics
+}
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -62,7 +62,6 @@ AND NOT protoPayload.serviceName="sts.googleapis.com"
 }
 
 resource "google_logging_project_bucket_config" "project" {
-  count            = var.log_analytics ? 1 : 0
   project          = local.project_id
   location         = "global"
   bucket_id        = "_Default"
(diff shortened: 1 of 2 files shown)


jasonthomas:
Since the logging bucket config is already provisioned automatically for projects for _Default and _Required buckets, (as per https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/logging_project_bucket_config), in your testing, does this require that the existing resource to be imported into state?

jasonthomas:
If the resource can just be created without an import (that is, we don't need to reconcile any config differences), I'd say we should just create `google_logging_project_bucket_config` resource without the condition on if `log_analytics` is enabled.

Flipez:
I was able to run tf and apply the changes to the bucket without reimporting a resource. If I change behavior as requested by you it will apply what looks like a default config like so to every project

```
  # module.prod_project["whattrainisitnow"].google_logging_project_bucket_config.project will be created
  + resource "google_logging_project_bucket_config" "project" {
      + bucket_id        = "_Default"
      + description      = (known after apply)
      + enable_analytics = false
      + id               = (known after apply)
      + lifecycle_state  = (known after apply)
      + location         = "global"
      + name             = (known after apply)
      + project          = "moz-fx-whattrainisitnow-prod"
      + retention_days   = 30
    }
```

Is that what you mean?

jasonthomas:
Yes, that's what I mean. Thanks for testing.

google_project/main.tf
```diff
@@ -60,3 +60,11 @@ AND NOT protoPayload.serviceName="iam.googleapis.com"
 AND NOT protoPayload.serviceName="sts.googleapis.com"
   EOT
 }
+
+resource "google_logging_project_bucket_config" "project" {
+  count            = var.log_analytics ? 1 : 0
```
jasonthomas, line 65: Per the previous comment, let's remove this so we define this resource for all projects.

jasonthomas (changes requested):

google_project/main.tf
jasonthomas, line 65:
```suggestion

```

jasonthomas (approved):

PR #144, opened 2023-12-04
feat(postgres): Set edition to enterprise by default

bkochendorfer:
Edition started to disappear from my changes over in https://github.com/mozilla-it/webservices-infra/pull/1291#issuecomment-1839227001


jbuck:
I found https://github.com/hashicorp/terraform-provider-google/issues/15553 which suggests this is a provider bug fixed by upgrading to [v5.3.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.3.0) or higher. I see fakespot is [using v4](https://github.com/mozilla-it/webservices-infra/blob/887a14452511011ac303fc94dfe49d6fc8c3baf5/fakespot/tf/prod/versions.tf#L8)

bkochendorfer:
oh okay, let's see!

bkochendorfer:
I think that seems to do it, thanks @jbuck !!

PR #145, opened 2023-12-05
feat(mozilla workgroup): OPST-1098 - Migrate internal workgroup to terraform-modules

tcotav:
We maintain a shim workgroup module in the *-infra repos to make it easier to reference workgroups. This means that this code is duplicated in several repos, and there is now drift. As part of consolidating permissions process and module, we will migrate workgroup module from the old location into this common one.

### DIFF
--- a/mozilla_workgroup/main.tf
+++ b/mozilla_workgroup/main.tf
@@ -1,5 +1,5 @@
 module "workgroup" {
-  source = "github.com/mozilla/terraform-modules//google_workgroup?ref=main"
+  source = "../google_workgroup"
   ids                           = var.ids
   roles                         = {}
   workgroup_outputs             = ["members", "google_groups"]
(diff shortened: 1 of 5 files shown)


mozilla_workgroup/main.tf
```diff
@@ -0,0 +1,8 @@
+module "workgroup" {
+  source = "github.com/mozilla/terraform-modules//google_workgroup?ref=main"
```
jasonthomas, line 2: Looks like this failing CI, need to run terraform lint here.

jasonthomas:
We are using semantic-release for versioning, so the commit message or the issue needs to follow https://github.com/mozilla/terraform-modules#versioning-and-releases

jasonthomas (approved):
LGTM!

PR #146, opened 2023-12-06
feat: OPST-1096 - terraform module to manage tenant roles #146

tcotav:
Provide a consistent way to manage GCP permissions for tenant GCP projects. 

Today it’s a lot of copy/pasta when generating new projects, but when permissions are added/changed, it requires a lot of replacements.


google_permissions/examples/admin_only/main.tf
```diff
@@ -0,0 +1,9 @@
+module "permissions" {
+  source = "../../../google_permissions"
+  # it is assumed that you loaded and have available a local.project
+  google_folder_id  = local.project.folder.id
+  google_prod_id    = local.project["prod"].id
```
jasonthomas, line 5:
```suggestion
  google_project_prod_id    = local.project["prod"].id
```

google_permissions/examples/admin_only/main.tf
```diff
@@ -0,0 +1,9 @@
...
+  google_nonprod_id = local.project["nonprod"].id
```
jasonthomas, line 6:
```suggestion
  google_nonprod_project_id = local.project["nonprod"].id
```

google_permissions/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+variable "admin_only" {
+  default     = false
+  description = "Whether or not to create a project with admin-only permissions."
+  type        = bool
+}
+
+/*
+// The following are sets of member lists to add to your project
+*/
+variable "admin_members" {
```
jasonthomas, line 47:
```suggestion
variable "admin_ids" {
```

To be consistent with the inputs of the [mozilla_workgroup module](https://github.com/mozilla/terraform-modules/tree/main/mozilla_workgroup#inputs), this should be renamed to ids.

google_permissions/other_perms.tf
```diff
@@ -0,0 +1,127 @@
+/*
+// This file is used to add additional permissions to the project. The idea is that
+// there are a set of pre-cleared permissions available for use, and that we can
+// add them to the project as needed by adding them to the `other_permissions` list.
+*/
+
+//
+// key: bq_job_user
```
jasonthomas, line 8: Style nit: There are a couple different styles of comments used here, especially between two options available for single line comments (`#`, `//`). For consistency I'd suggest using one.

google_permissions/examples/basic_with_addon/main.tf
```diff
@@ -0,0 +1,14 @@
+module "permissions" {
+  source = "../../../google_permissions"
+  # it is assumed that you loaded and have available a local.project
+  google_folder_id  = local.project.folder.id
+  google_prod_id    = local.project["prod"].id
+  google_nonprod_id = local.project["nonprod"].id
+  admin_members     = ["workgroup:my-project/admins"]
+  developer_members = ["workgroup:my-project/developers"]
+  other_permissions = [
```
jasonthomas, line 9:
```suggestion
  other_permissions = [
```

Let's discuss how we want to expose this to users later today. I feel like we should use the actual role names instead of these stubbed names to reduce confusion. Additionally, we should should have some feedback (maybe via terraform validation conditions) for non supported roles, document supported roles, or both.

jasonthomas, line 9:
```suggestion
  nonprod_roles = [
  prod_roles = [
```

tcotav: @jasonthomas  - one thing first - I created two distinct sets of allowed roles in case we want specific targeting different environments. Not sure if necessary, but seemed a good idea?

Second, did some homework on the solution. I'll go with option 3, but wanted to run it by you first.

```hcl
# see https://github.com/hashicorp/terraform/issues/25609#issuecomment-1472119672

# Three ways to do the validation of a list of roles against an allowed list


locals {
  # This is a list of all the permissions that we allow to be added to the project.
  # If you want to add a new permission, add it to this list.
  prod_allowed_roles = [
    "roles/bigquery.jobUser",
    "roles/automl.editor",
    "roles/cloudtranslate.editor",
    "roles/storage.objectAdmin",
    "roles/translationhub.admin",
    "roles/storage.admin",
    "roles/cloudsql.admin"
  ]

  nonprod_allowed_roles = [
    "roles/bigquery.jobUser",
    "roles/editor",
    "roles/storage.admin",
    "roles/cloudsql.admin"
  ]
}

variable "prod_roles" {
  description = "List of roles to apply to the prod project."
  type        = list(string)
  default     = []

  # WAY #1 - don't use locals -- have it all here, BUT then have to maintain list of roles in this block
  validation {
    condition = alltrue([
      for x in var.prod_roles : contains([
        "roles/bigquery.jobUser",
        "roles/automl.editor",
        "roles/cloudtranslate.editor",
        "roles/storage.objectAdmin",
        "roles/translationhub.admin",
        "roles/storage.admin",
        "roles/cloudsql.admin"
      ], x)
    ])
    error_message = "You have specified an invalid prod role."
  }
}

# WAY #2 - locals with dummy variable. throws kinda sketchy error
locals {
  validate_prod_roles = alltrue([
    for x in var.prod_roles : contains(local.prod_allowed_roles, x)
  ]) ? true: tobool("You have specified an invalid prod role.")
}

variable "nonprod_roles" {
  description = "List of roles to apply to the nonprod project."
  type        = list(string)
  default     = []
  }

}

# WAY #3 - recommended in the bug above, same kind of sketchy error though
output "validate_nonprod_roles" {
  value = null

  precondition {
    condition = alltrue([
      for x in var.nonprod_roles : contains(local.nonprod_allowed_roles, x)
    ])
    error_message = "You have specified an invalid nonprod role."
  }
}

##################################################  
#
# Here's the way the error looks for the third way. Also similar in second. You get the error msg though.
# 
#Plan: 6 to add, 0 to change, 5 to destroy.
#╷
#│ Error: Module output value precondition failed
#│ 
#│   on ../../../../../mozilla/terraform-modules/google_permissions/variables.tf line 82, in output "validate_nonprod_roles":
#│   82:     condition = alltrue([
#│   83:       for x in var.nonprod_roles : contains(local.nonprod_allowed_roles, x)
#│   84:     ])
#│     ├────────────────
#│     │ local.nonprod_allowed_roles is tuple with 4 elements
#│     │ var.nonprod_roles is list of string with 2 elements
#│ 
#│ You have specified an invalid nonprod role.
```

tcotav: re: nonprod_roles + prod_roles question above -- there is also folder level roles which is not captured

jasonthomas: Using precondition seems like a reasonable approach, but this could also be used with a resource definition instead of the output like:

```
# we create one resources for all nonprod project role bindings, and enable/disable on admin_only variable
# what's nice about this approach is the precondition is not evaluated when admin_only is true. 
resource "google_project_iam_binding" "nonprod_roles" {
  for_each = toset([for role in var.nonprod_roles : x if !var.admin_only])
  project  = var.google_nonprod_id
  role     = each.value
  members  = module.developers_workgroup.members

  lifecycle {
    precondition {
      condition = alltrue([
        for x in var.nonprod_roles : contains(local.nonprod_allowed_roles, x)
      ])
      error_message = "You have specified an invalid nonprod role."
    }
  }
}
```

jasonthomas: But also noisy because precondition will fail for each resource attempting to be created.

google_permissions/variables.tf
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// roles that are intended for the non-production project are in this list
+variable "nonprod_roles" {
+  description = "List of roles to apply to the nonprod project."
+  type        = list(string)
+  default     = []
+}
+
+output "validate_nonprod_roles" {
```
jasonthomas, line 89: Style: Move outputs out of variable.tf into outputs.tf

google_permissions/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 3.0"
+    }
+  }
+  required_version = "~> 1.0"
```
jasonthomas, line 8:
```suggestion
  required_version = ">= 1.2"
```
Since precondition is only available in 1.2.

google_permissions/examples/basic/main.tf
```diff
@@ -0,0 +1,9 @@
+module "permissions" {
+  source = "../../../google_permissions"
+  // it is assumed that you loaded and have available a local.project
+  google_folder_id          = local.project.folder.id
+  google_prod_project_id    = local.project["prod"].id
+  google_nonprod_project_id = local.project["nonprod"].id
+  admin_ids                 = ["workgroup:my-project/admins"]
+  developer_ids             = ["workgroup:my-project/developers"]
```
jasonthomas, line 8:
```suggestion
  developer_ids             = ["workgroup:my-project/developers"]
  viewer_ids                   = ["workgroup:my-project/viewers"]
```

jasonthomas: Include viewers as part of basic example

google_permissions/examples/basic/main.tf
```diff
@@ -0,0 +1,9 @@
+module "permissions" {
+  source = "../../../google_permissions"
+  // it is assumed that you loaded and have available a local.project
+  google_folder_id          = local.project.folder.id
+  google_prod_project_id    = local.project["prod"].id
+  google_nonprod_project_id = local.project["nonprod"].id
+  admin_ids                 = ["workgroup:my-project/admins"]
```
jasonthomas, line 7:
```suggestion
  admin_ids                 = ["workgroup:workgroup_name/workgroup_subgroup"]
```

google_permissions/main.tf
```diff
@@ -0,0 +1,56 @@
...
+  folder  = var.google_folder_id
+  role    = "roles/owner"
+  members = module.admins_workgroup.members
+}
+
+resource "google_folder_iam_binding" "viewer" {
+  count   = var.admin_only ? 0 : 1
+  folder  = var.google_folder_id
+  role    = "roles/viewer"
+  members = module.developers_workgroup.members
```
jasonthomas, line 13:
```suggestion
  members = toset(
  module.developers_workgroup.members,
  module.viewers_workgroup.members
  )
```

jasonthomas:
It looks like Semantic PR is sad, the title of the PR needs to be adjusted.

jasonthomas (approved):

PR #148, opened 2024-01-05
Circleci OIDC

whd:
Companion to https://github.com/mozilla-it/global-platform-admin/pull/1021. The changes here are backwards compatible with the existing module (count=1 counts as an internal state change and is otherwise a no-op). The assertions for CircleCI are somewhat bespoke to our specific mapping (though generally follow CircleCI's recommendation where possible) so this isn't necessarily a generic module usable outside of Mozilla.

It's possible to create an entirely separate module instead of using the one for github actions. 

https://github.com/mozilla-services/cloudops-infra/compare/DSRE-1261?expand=1 contains a draft of how this module might be used to convert some of our existing data workloads from using static credentials in CI to OIDC.


google_deployment_accounts/variables.tf
```diff
@@ -20,6 +20,54 @@ variable "gha_environments" {
   default     = []
 }
 
+# FIXME consider breaking this out into multiple variables
```
whd, line 23: I think I prefer specifying a single variable to keep the terraform interface simple but it means module users need to build these strings themselves. Using multiple variables instead maps more closely to how we are doing this for github actions, it's just that we happen to only have one variable `input_gha_actions`  on the gha side. I think we did a good job of avoiding the complexity of https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#example-subject-claims by requiring environments in GHA but this isn't really viable for CircleCI.

 I'll likely remove the commit with the commented out version unless somebody has a preference to enumerate each assertion as a variable.

google_deployment_accounts/variables.tf
```diff
@@ -1,11 +1,17 @@
 variable "account_id" {
   type        = string
-  description = "Name of the service account. Defaults to deploy-<env>"
+  description = "Name of the service account. Defaults to deploy-ENV."
```
whd, line 3: I changed the formattinng here because `<env>` appears to get dropped in markdown conversion.

whd:
@jbuck this is ready for review, and I've included some examples as well. I think the default here might be a little dangerous compared to what we have for GHA since environment isn't a concept for CircleCI. We could consider making the default only use the `main` branch but then we'd then need to figure out how to support all branches explicitly.

https://github.com/mozilla-it/dataservices-infra/pull/137 was planned from this branch at the time of this comment, hopefully demonstrating that this is backwards compatible with all the existing GHA use cases.

https://github.com/mozilla-services/cloudops-infra/pull/5387 contains the real-world examples I'm hoping to use this with.

jbuck (approved):

PR #149, opened 2024-01-16
feat: Google GKE Add ability to enable gateway api 

bkochendorfer:
Default to standard channel


bkochendorfer:
I'm not sure what is best here, to pass in an `enabled = true` instead and default to `STANDARD` channel or use this and have the ability to pass in different channels.

jasonthomas:
> I'm not sure what is best here, to pass in an `enabled = true` instead and default to `STANDARD` channel

I don't think we'd use the `EXPERIMENTAL` channel unless it's for testing purposes, and not sure if it would be easy to switch between them on an existing cluster. Being opinionated and simplifying to default to `STANDARD` channel seems reasonable to me.

bkochendorfer:
Agreed, thanks for the feedback @jasonthomas

jasonthomas (approved):

PR #150, opened 2024-01-31
feat(deployment_accounts): Add support for multiple repositories

Flipez:


ahoneiser (approved):

jasonthomas (dismissed):

google_deployment_accounts/variables.tf
```diff
@@ -40,3 +40,9 @@ variable "github_repository" {
   type        = string
   description = "The Github repository running the deployment workflows in the format org/repository"
 }
+
+variable "github_repositories" {
+  type        = list(string)
+  description = "The Github repositories running the deployment workflows in the format org/repository, overwriting the github_deployment variable"
```
whd, line 46:
```suggestion
  description = "The Github repositories running the deployment workflows in the format org/repository, overwriting the github_repository variable"
```

google_deployment_accounts/variables.tf
```diff
@@ -40,3 +40,9 @@ variable "github_repository" {
   type        = string
   description = "The Github repository running the deployment workflows in the format org/repository"
```
whd, line 41: Note that this variable still appears to be required. I was planning to make it optional in my PR: https://github.com/mozilla/terraform-modules/pull/148/files#diff-cf11d21cae1e82df3acc59244bf503b8992e1d0bf9590b2b8ad0ee5e2f5f44adR93. Making it optional makes the it possible to specify neither `github_repositories` nor `github_repository` and have the `coalescelist` fail but that seems like an acceptable and obvious error condition to module users.

Another option would to continue to require one github repository be specified this way and rename `github_repository` to e.g. `additional_github_repositories`.

whd (commented):
I have an open PR https://github.com/mozilla/terraform-modules/pull/148 that's vaguely related to this in that it is providing an interface for specifying OIDC access across a couple of different assertion types (CircleCI side), a subset of which would be supporting access from multiple GH repos. This PR doesn't clash with that one but it's worth noting that with GHA we are very opinionated about requiring `environment`(s). With this PR we're going to require that the same environments be used across all the repositories specified. In the case where we'd want some per-respository environments in GHA we'd need to add something similar to [circleci_attribute_specifiers](https://github.com/mozilla/terraform-modules/blob/d37ea540adbc371d9bdc2dcad2b4428fb86d50a1/google_deployment_accounts/README.md#input_circleci_attribute_specifiers).

Flipez:
@whd I would go ahead and merge that PR because it's blocking us on a migration but I think your points are valid. But maybe, opposed to add specific variables for circleci we could rewrite the handling in general (and introduce a possible breaking change) to make it more generalised?

ahoneiser (approved):

whd:
> I would go ahead and merge that PR because it's blocking us on a migration but I think your points are valid. 

By "that PR", are you referring to this PR or https://github.com/mozilla/terraform-modules/pull/148? I don't have any objection to merging this PR as-is.

> But maybe, opposed to add specific variables for circleci we could rewrite the handling in general (and introduce a possible breaking change) to make it more generalised?

This is a reasonable idea, but difficult to do without increasing the amount of cognitive load for users of the module since you'd need to understand a bunch of information about OIDC assertions and attribute mappings that differ between providers ([GHA](https://github.com/mozilla-it/global-platform-admin/blob/main/global-shared/tf/workload-identity.tf#L14) vs [CircleCI](https://github.com/mozilla-it/global-platform-admin/blob/main/global-shared/tf/workload-identity.tf#L57)). Given that you must specify `wip_name` in module configuration anyway (indeed, a repo by itself could have both GHA and CircleCI configured) I think it's best to be backwards compatible with the existing interface while still providing some flexibility with CircleCI. I consider it a useful feature that we are very opinionated about the GHA assertion flows and the complexity of the CircleCI stuff is only because we can't be that opinionated when retrofitting onto existing CI configuration.

The above of course identifies that there's an edge case where you might want to grant access to both identify pools. The interface I'm recommending could be extended in a backwards compatible way by `wip_name` but has the same issue with `github_repository` when adding `github_repositories`. Ultimately the purpose of this module is to make things simpler and easier to configure for the "happy path" cases and I think that users can fall back to managing the assertions themselves outside of the module if necessary.

jasonthomas (approved):
Let's land this PR, and address the other changes in a followup PR.

Flipez:
> By "that PR", are you referring to this PR or https://github.com/mozilla/terraform-modules/pull/148? I don't have any objection to merging this PR as-is.

Yes I mean this PR (#150)

I understand that it is not that easy to generalize the module. I have no strong opinions and are also fine with keeping it less generalized. Possibly also easier to identify room for improvements after the features are in use.

google_deployment_accounts/variables.tf
whd: This was fixed in https://github.com/mozilla/terraform-modules/pull/152

PR #167, opened 2024-04-15
feat(google_gke): Allow enabling container streaming on a node pool

bkochendorfer:
Allows enabling Google Container Filesystem for image stream https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#gcfs_config

This will require re-creating a node pool so should not be enabled by default.


google_gke/variables.tf
```diff
@@ -231,6 +231,14 @@ variable "node_pools_guest_accelerator" {
   }
 }
 
+variable "node_pools_enable_gcfs" {
+  description = "Map containing node pools guest accelerator. Each node pool's name is the key. See locals.tf for defaults."
```
jasonthomas, line 235:
```suggestion
  description = "Map containing node pools to enable Google Container File System (gcfs) image streaming."
```

jasonthomas (approved):
r+wc

jasonthomas (approved):

jasonthomas (approved):

PR #172, opened 2024-04-18
feat(google_gke): enable internal ip reservation for k8s api proxy

dlactin:
Added the option to reserve a static internal IP address for each cluster. 

This is intended for the k8s api proxy service running in each cluster but could be used elsewhere.

### DIFF
--- a/google_gke/k8s_api_proxy.tf
+++ b/google_gke/k8s_api_proxy.tf
@@ -1,0 +1,13 @@
+#
+# K8S API Proxy Setup
+#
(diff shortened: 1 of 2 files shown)


bkochendorfer (approved):
Seems fine to me, how is the IP attached to the proxy?

dlactin:
> Seems fine to me, how is the IP attached to the proxy?

The IP will be consumed via a k8s-api-proxy chart deployed during bootstrapping, working on that component now. 

Essentially a simple nginx deployment with an internal lb.

PR #180, opened 2024-05-13
feat(google_monitoring): add module

ahoneiser:


jasonthomas (approved):

smarnach (commented):
It would be nice to expose `period` and `user_labels` as input variables as well.

ahoneiser:
thanks for the feedback @smarnach, added `period` and `user_labels`

biancadanforth:
Thanks for working on this -- and I independently added the two options Sven suggested in my own PR last night, so 100% agree on those additions: https://github.com/mozilla-it/webservices-infra/pull/2279.

One thing I was going to add to my own PR are the `google_monitoring_notification_channel` and `google_monitoring_alert_policy`. Should those be a part of this module, or part of some companion module, or at least referenced in the README for this module? I suppose not everyone will want alerts with notifications for their checks, but then again, uptime is one of those golden metrics that we're really focusing on.

google_monitoring/uptime_checks.tf
```diff
@@ -0,0 +1,24 @@
+resource "google_monitoring_uptime_check_config" "https" {
+  for_each = { for uptime_check in var.uptime_checks : uptime_check.name => uptime_check }
```
smarnach, line 2: I'm not convinced it's useful to have this loop inside the module, rather than using
```
module "uptime_check" {
    source = "..."
    for_each = {...}
    ...
}
```
when instantiating the module and you want more than one check. I'd personally prefer to do loops when instatntiating the module. I think that's more flexible, since it makes it easier to share, say, the `user_labels` across all your checks.

(It probably doesn't matter much either way.)

smarnach:
> Edit: Actually, the idea (which aligns with some examples you shared elsewhere @ahoneiser) is to create the alerts/notifications in Grafana as opposed to in Google Monitoring. Is there Terraform config for that as well? I didn't see it in the `remote-settings/tf` directory.

We don't use Terraform for Grafana at all. It would be nice to do that at some point, but it requires a serious amount of work, in particular if we want to use it for dashboards.

google_monitoring/uptime_checks.tf
biancadanforth: Thinking about this from the `symbols` project perspective, where we have two services sharing the same GCP project, would this module-level approach prevent us from having different user labels for the different services?

smarnach: Either approach would allow having different labels for different checks. As I said, it probably doesn't matter much either way. I just think it's a bit weird to have that loop inside the module. We don't do that for the database modules, so why should we do it here? If someone wants a loop, they can always add to the instantiation.

ahoneiser: fair point, this is currently based on the idea of the module calling project passing the checks in as a list, module takes care of working the details out .. I'm thinking about passing this in from e.g. a yaml file or something like this (that's easy to read and update by clients), working with it can still be done via locals easily .. removing the loop from the module might make it less complex, and easier for the users to maintain :thinking:

ahoneiser: I love the flexibility of the `optional` keyword in objects a lot, so personally always going for that .. also meant this to eventually become something that could also include other things like uptime_checks, so wanted to have a single list / map referring to the uptime checks that should be created :thinking:

ahoneiser: if it would remain to be only creating uptime_checks, I could name it differently and make the uptime_check relevant vars top level vars :thinking:

PR #193, opened 2024-06-07
fix(google_redis): Remove ignore_changes on maintenance_schedule

jbuck:


bqbn (approved):
I've been wanting to do this. Thanks!

PR #203, opened 2024-06-26
chore(google_gke): Allow using google and google-beta v5 providers

jbuck:


google_gke/cluster.tf
```diff
@@ -289,7 +284,15 @@ resource "google_container_node_pool" "pools" {
...
-    taint = local.node_pools_taints[each.key]
+    dynamic "taint" {
+      for_each = local.node_pools_taints
+
+      content {
+        key    = each.value.key
+        value  = each.value.value
+        effect = each.value.effect
+      }
+    }
```
bkochendorfer, line 295: I think as I mentioned I had to use a lifecycle setting here when testing locally but let's give it a shot and see what happens.

bkochendorfer (approved):

jbuck:
I did some testing with the sandbox low cluster, and it appears to be working correctly, just need to do the state magic when adding it to each cluster https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/version_5_upgrade#node_config0taint-and-node_pool0node_config0taint-field-change

bkochendorfer (approved):

PR #205, opened 2024-06-26
Opst 1260 birdbox additional roles

Basma1912:


Basma1912:
the changes are made because the birdbox permissions.tf is using them 
https://github.com/mozilla-it/webservices-infra/blob/main/birdbox/tf/global/permissions.tf

ahoneiser (approved):

PR #208, opened 2024-07-01
feat(google_monitoring): add more options to module

ahoneiser:


cvalaas (approved):

google_monitoring/uptime_checks.tf
```diff
@@ -7,10 +7,23 @@ resource "google_monitoring_uptime_check_config" "https" {
...
-    use_ssl      = true
-    validate_ssl = true
+    path                = each.value.path
+    port                = 443
+    request_method      = each.value.request_method
+    use_ssl             = true
+    validate_ssl        = true
+    content_type        = lookup(each.value, "content_type", null)
+    custom_content_type = lookup(each.value, "custom_content_type", null)
+    body                = lookup(each.value, "body", null)
```
biancadanforth, line 17: Do you know that `null` is a valid value to be passed in here as the default? For example, `content_type` [per the docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_uptime_check_config#content_type) says possible values are `TYPE_UNSPECIFIED`, `URL_ENCODED`, `USER_PROVIDED`, so we'd probably want `TYPE_UNSPECIFIED` here as the default?

Less clear what the default should be for `custom_content_type` and `body` when not provided.

google_monitoring/uptime_checks.tf
```diff
@@ -7,10 +7,23 @@ resource "google_monitoring_uptime_check_config" "https" {
...
+  }
+
+  dynamic "content_matchers" {
+    for_each = each.value.content_matchers
+
+    content {
+      content = content_matchers.value.content
+      matcher = content_matchers.value.matcher
+    }
```
biancadanforth, line 26: Same situation here -- except the [`content_matchers` docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_uptime_check_config#content_matchers) explicitly state that:

> This field is optional and should only be specified if a content match is required.

biancadanforth (commented):
Thanks Andre for trying to make this module more customizable. I have some concerns about the defaults for these optional properties, which I mention inline. Have you tried this module out on a basic check that doesn't use these new optional properties?

I can try on mine, but I probably won't get to it until later today or tomorrow best case scenario.

**Edit: I had a chance to try this out on my [PR](https://github.com/mozilla-it/webservices-infra/pull/2279), and I found it to work just fine without using any of these additional optional arguments, so LGTM!**

google_monitoring/uptime_checks.tf
biancadanforth: Ah okay, [variables are `nullable: true` by default](https://developer.hashicorp.com/terraform/language/values/variables#disallowing-null-input-values), also:

> If you set an argument of a resource to null, Terraform behaves as though you had completely omitted it...

So that will work just fine, and I've confirmed this by updating my [PR](https://github.com/mozilla-it/webservices-infra/pull/2279) and `atlantis plan` says no changes.

google_monitoring/uptime_checks.tf
biancadanforth: Now that I've had more time to try this out and inspect the code, I see you default `content_matchers` to an empty list, `[]`. So I imagine that means we never pass this argument in either as a result. The fact that my uptime check PR using this config is still working after `atlantis apply` supports that.

PR #211, opened 2024-07-09
feat(google_cdn_backend_bucket): add bypass_cache_on_request_headers block to cdn_policy input

biancadanforth:
Because:
* Tecken needs a CDN with no caching in front of its public GCS bucket.
* Part of the config to disable caching for the CDN is not currently exposed on this module.

This commit:
* Updates the `cdn_policy` input to the module to pass an optional `bypass_cache_on_request_headers` key.

### DIFF
--- a/google_cdn_backend_bucket/main.tf
+++ b/google_cdn_backend_bucket/main.tf
@@ -18,6 +18,12 @@ resource "google_compute_backend_bucket" "default" {
     max_ttl           = var.cdn_policy.max_ttl
     negative_caching  = var.cdn_policy.negative_caching
     serve_while_stale = var.cdn_policy.serve_while_stale
+    dynamic "bypass_cache_on_request_headers" {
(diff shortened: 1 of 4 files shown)


google_cdn_backend_bucket/main.tf
```diff
@@ -18,6 +18,9 @@ resource "google_compute_backend_bucket" "default" {
     max_ttl           = var.cdn_policy.max_ttl
     negative_caching  = var.cdn_policy.negative_caching
     serve_while_stale = var.cdn_policy.serve_while_stale
+    bypass_cache_on_request_headers {
+      header_name = var.cdn_policy.bypass_cache_on_request_headers.header_name
+    }
```
biancadanforth, line 23: This isn't exactly what I want -- as written, it will set the `bypass_cache_on_request_headers` block with a default `"Cache-Control"` value for `header_name`. I don't want to set the `bypass_cache_on_request_headers` block at all if its not provided.

PR #220, opened 2024-08-30
feat(google_project): Include data access logs for additional Google Cloud services

jbuck:


jasonthomas (dismissed):

google_project/locals.tf
```diff
@@ -32,4 +32,7 @@ locals {
     "stackdriver.googleapis.com",
   ]
   all_project_services = setunion(local.default_project_services, var.project_services)
+
+  default_data_access_logs = ["iam.googleapis.com", "secretmanager.googleapis.com", "sts.googleapis.com"]
```
jmhodges, line 36: Not binding and out of curiosity: Do we also need cloudkms to get the key accesses under secretmanager for things like the hsm slot updates and such when autograph is doing that work? (I'm happy for us to do that later when we actually need it)

jasonthomas (approved):

PR #223, opened 2024-09-03
fix(google_cloudsql_postgres): Allow null values for connector_enforcement.

smarnach:
This issue prevented me from deploying: https://github.com/mozilla-it/webservices-infra/pull/2654#issuecomment-2326543360

In Terraform, the logical or operator `||` does not short-circuit. As a result, the second part of the validation condition gets evaluated even when `connector_enforcement` is `null`, and for some reason that's not a valid second parameter to `contains()` (which makes no sense, but it is how it is).


relud (approved):

jbuck (approved):

jbuck:
ack, sorry @smarnach! Thank you for fixing this

jbuck:
We'll need the same fix in the mysql module too, bleh

PR #225, opened 2024-09-05
feat(google_monitoring): add selected_regions option to uptime check

biancadanforth:
As part of [OBS-191](https://mozilla-hub.atlassian.net/browse/OBS-191), I am writing a ["How to create an uptime check"](https://docs.google.com/document/d/1Q9BW41-hhHwt03uOsm4oKYUoJdmsxPGkPrpOcwmuNco/edit?usp=sharing) document.

One common use case we expect is for teams to only care about checks for a subset of all the global regions a Google Uptime Check checks from by default (e.g. maybe they only need checks from `"USA"` and `"EUROPE"`, and not from `"SOUTH_AMERICA"` or `"ASIA_PACIFIC"`). Since each checker costs us executions, it makes sense to have this option, so I've added it here.

I've tested it in https://github.com/mozilla-it/webservices-infra/pull/2698.

### DIFF
--- a/google_monitoring/README.md
+++ b/google_monitoring/README.md
@@ -1,4 +1,3 @@
-<!-- BEGIN_TF_DOCS -->
 ## Requirements
 
 | Name | Version |
@@ -32,4 +31,3 @@ No modules.
 ## Outputs
 
 No outputs.
-<!-- END_TF_DOCS -->
--- a/google_monitoring/README.md
+++ b/google_monitoring/README.md
@@ -1,3 +1,4 @@
+<!-- BEGIN_TF_DOCS -->
 ## Requirements
 
 | Name | Version |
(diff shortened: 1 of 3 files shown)


google_monitoring/README.md
```diff
@@ -1,3 +1,4 @@
+<!-- BEGIN_TF_DOCS -->
```
ahoneiser, line 1: piping to the readme like `$ terraform-docs markdown . |tee README.md` will not add tags to the `README.md` file :slightly_smiling_face:

google_monitoring/variables.tf
```diff
@@ -14,6 +14,7 @@ variable "uptime_checks" {
     timeout             = optional(string, "60s")
     period              = optional(string, "300s")
     user_labels         = optional(map(string), {})
+    selected_regions    = optional(list(string), [])
```
ahoneiser, line 17: I think we should add defaults here. I've been adding this to the refractr monitoring recently, google uptime checks require at least 3 regions to work, so I think adding those as defaults would be a good idea, e.g.:

```
selected_regions = optional(list(string), ["EUROPE", "US_OREGON", "US_VIRGINIA"])
```

Also, looking at future costs, 3 regions means 3 check executions, leaving this with the default means 6 check executions, so also a good idea to keep this low in defaults :thinking:

bkochendorfer (dismissed):
Once Andre's suggestions are pulled in looks good to me

bkochendorfer (approved):

ahoneiser (approved):

PR #234, opened 2024-11-06
chore(OPST-1385): renaming google workgroups module to mozilla workgroup

Basma1912:


bkochendorfer (commented):
Probably no need to keep the old one around as `mozilla_workgroup_old`, nothing should call that right?

Basma1912:
that is correct -- I renamed it because I was too lazy to delete it .

mozilla_workgroup/README.md
```diff
@@ -1,76 +1,50 @@
-# Mozilla Workgroup
+# workgroup
```
jasonthomas, line 1:
```suggestion
# Mozilla workgroup
```

mozilla_workgroup/README.md
```diff
@@ -1,76 +1,50 @@
...
-| <a name="module_workgroup"></a> [workgroup](#module\_workgroup) | github.com/mozilla/terraform-modules//google_workgroup | main |
-
-## Resources
-
-No resources.
+This module is cloned from https://github.com/mozilla-services/cloudops-infra-terraform-modules/tree/master/data-workgroup.
+## Example
+```hcl
+module "workgroup" {
+  source = "github.com/mozilla/terraform-modules//google_workgroup?ref=main"
```
jasonthomas, line 24:
```suggestion
  source = "github.com/mozilla/terraform-modules//google_workgroup?ref=main"
```

This should be mozilla_workgroup reference, but this is pulled in from the example1.tf, so that needs to be updated first and then the doc regenerated.

mozilla_workgroup/variables.tf
```diff
@@ -7,3 +7,32 @@ variable "ids" {
...
 }
+
+variable "roles" {
+  type        = map(string)
+  description = "List of roles to generate bigquery acls for"
+  default = {
+    metadata_viewer = "roles/bigquery.metadataViewer"
+    read            = "READER"
+    write           = "WRITER"
+  }
```
jasonthomas, line 18:
```suggestion
  default = {}
```
I am of the opinion to change this to an empty map, and put a comment of potential options.

Basma1912: please take a look at the comment I added and let me know .. thanks

mozilla_workgroup/README.md
jasonthomas: It looks like this was updated on the README directly, but this actually pulled in from the example via terraform-docs command using this reference https://github.com/mozilla/terraform-modules/blob/d8c958ff41991450d897ca6cd40ab25fdda87bb5/mozilla_workgroup/.terraform-docs.yml . 

We should update https://github.com/mozilla/terraform-modules/blob/d8c958ff41991450d897ca6cd40ab25fdda87bb5/mozilla_workgroup/examples/example1.tf. 

Once example1.tf is updated, run `terraform-docs` to generate the README document.

Basma1912: oh I accepted the changes on the README .. I will update the example1.tf

mozilla_workgroup/variables.tf
jasonthomas: LGTM

jasonthomas (approved):

PR #236, opened 2024-12-03
feat:add aws_gke_oidc_config and aws_gke_oidc_role modules

amitchell-moz:
Add modules to provision AWS roles + OIDC configs to allow GKE workloads to assume AWS roles.

https://mozilla-hub.atlassian.net/browse/OPST-1509

This PR introduces 2 modules:
- aws_gke_oidc_config
  - creates an OIDC provider in AWS to setup a trust relationship between a cluster & AWS account
- aws_gke_oidc_role
  - creates AWS roles that use the trust relationship established by ^^ to allow GKE service accounts to assume AWS roles

These need to be separate modules because the OIDC provider URL must be unique per-account, but a given GKE cluster only has 1 OIDC endpoint. That means the OIDC provider tf must only be ran once, after that any number of roles can use it.

## Changelog entry
```
Adds aws_gke_oidc_config and aws_gke_oidc_role modules
```


aws_gke_oidc_role/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+}
+
+variable "gke_service_account" {
+  description = "GKE service account to grant role assumption privilleges"
+  type        = string
+}
+
+### Optional
+
+variable "tags" {
```
jbuck, line 45: I think you can remove this and add tags at the provider level with [`default_tags`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-1)

jbuck (approved):
Could you delete the `.terraform.lock.hcl` files? This looks good to me!

amitchell-moz:
Rebasing the new CI stuff in here blew up - opening a new clean PR

PR #244, opened 2025-01-06
fix(module):google_project_module_update_skip_delete_arg

Basma1912:
## Changelog entry
```
The Argument skip_delete is deprecated and will be removed in 6.0.0. 
and is replaced by the argument deletion_policy 
```


bkochendorfer (approved):
Looks to be the same result

Basma1912:
yes, it is the same result , just updating the argument.

PR #247, opened 2025-01-07
Opst 1770 

Basma1912:
## Changelog entry
```
change the value of the deletion_policy to PREVENT if the project name contains Bastion 
```

### DIFF
--- a/google_project/main.tf
+++ b/google_project/main.tf
@@ -10,7 +10,7 @@ resource "random_id" "project" {
 resource "google_project" "project" {
   name            = local.display_name
   project_id      = local.project_id
-  deletion_policy = "ABANDON"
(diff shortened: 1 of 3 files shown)


google_project/main.tf
```diff
@@ -10,7 +10,7 @@ resource "random_id" "project" {
 resource "google_project" "project" {
   name            = local.display_name
   project_id      = local.project_id
-  deletion_policy = "ABANDON"
+  deletion_policy = contains(local.project_id, "bastion") ? "PREVENT" : "ABANDON"
```
bkochendorfer, line 13: hmm, maybe instead we could pass in the `deletion_policy` as a variable but default to `ABANDON`?

Basma1912: ok I need to update the bastion main.tf when calling the module.

bkochendorfer (approved):

PR #253, opened 2025-01-29
feat(google_gke_namespace_logging): Output linked dataset ID

jbuck:
This is needed for teams needing to change the default permissions on
the linked BigQuery dataset.

## Changelog entry
```
feat(google_gke_namespace_logging): Output linked BigQuery dataset ID
```

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -168,6 +168,9 @@ runs:
         GH_TOKEN: ${{ github.token }}
       shell: bash
       run: |
+        # Workaround "error: insufficient permission for adding an object to repository database .git/objects"
+        # https://github.com/terraform-docs/gh-actions/issues/50#issuecomment-1231314373
+        sudo chown runner:docker -R .git
         git pull # In case we have multiple changelog updates
         FILE_TO_COMMIT="$PACKAGE_NAME/CHANGELOG.md"
         if ! git diff --quiet --exit-code $FILE_TO_COMMIT ; then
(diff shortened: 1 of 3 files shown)


whd (dismissed):
I had a discussion with @jbuck about this. related to https://console.cloud.google.com/support/cases/detail/v2/56963028?invt=AboJ0Q&project=moz-fx-autograph-nonprod

Basically we're compensating for a (arguably correct) behavior change on GCP's end in provisioning these logging datasets. Previously they were provisioned with projectReaders, projectOwners, projectEditors passed through to the dataset, but now they are not. One way to do this is to implement the old behavior in this module directly, possibly using a flag to disable that behavior. Another way is to pass the responsibility of managing permissions onto a downstream module, which is more flexible and what this PR is enabling.

Overall I'm not yet convinced it's worth passing the responsibility downstream. It ultimately boils down to https://cloud.google.com/logging/docs/analyze/query-linked-dataset#access-control-log-views. Though I am generally a staunch proponent of NOT using special groups with BQ [(see these rules)](https://github.com/mozilla-it/dataservices-infra/blob/main/.tflint.d/policies/bigquery.rego) I am more OK with using them if we do so intentionally and explicitly. 

Basically the question is this: do we want people with project-level Viewer access to have access to these logs as well? I imagine the answer is somewhat similar to asking whether Viewers should have private log viewer access (they don't), but we can grant our tenant developers this access and probably do.

amitchell-moz (approved):

PR #256, opened 2025-02-01
chore: Add terraform-docs delimiters

jbuck:
Add terraform-docs delimiters to all modules missing them so docs are generated without duplicates

## Changelog entry
```
Add terraform-docs delimiters
```

PR #257, opened 2025-02-01
fix(docs): Sign commits for documentation changes

jbuck:
Once again I need to change a lightbulb https://www.youtube.com/watch?v=AbSehcT19u0

Our story begins with 2 small PRs:
- https://github.com/mozilla/terraform-modules/pull/253
- https://github.com/mozilla/terraform-modules/pull/254

Between these two PRs I noticed that:
- I wasn't able to merge my PR, because it had some unsigned commits, and this repository requires signed commits
- The generated documentation was being duplicated by terraform-docs

The reason the generated documentation was being duplicated is because most of our README.md files are missing the block delimeters `<!-- BEGIN_TF_DOCS -->` and `<!-- END_TF_DOCS -->`.
So the generated content will be appended at the end of the file.
Well no worries, maybe I just need to add those blocks to our existing README.md files? https://github.com/mozilla/terraform-modules/pull/256
Close, but I think I need to add `.terraform-docs.yml` config file to every directory as well. And we need to sign the commits too! Alas

So this PR switches the README.md generation to use signed commits like #240 did.

## Changelog entry
```
fix(docs): Sign commits for documentation changes
```

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -99,14 +99,35 @@ runs:
         else
           echo "tf_docs=false" >> "$GITHUB_OUTPUT"
         fi
-    - name: Render terraform docs inside the README.md and push changes back to PR branch
+    - name: Render terraform docs inside the README.md
       if: ${{ steps.tf_docs.outputs.tf_docs == 'true' }}
+      id: terraform_docs
       uses: terraform-docs/gh-actions@v1.3.0
       with:
         working-dir: ${{ inputs.package-name }}
         output-file: README.md
-        git-push: "true"
         config-file: .terraform-docs.yml
+    - name: If documentation is updated, push to PR branch with a signed commit
+      if: ${{ steps.terraform_docs.outputs.num_changed != '0' }}
+      env:
+        DESTINATION_BRANCH: ${{ github.event.pull_request.head.ref }}
+        FILE_TO_COMMIT: ${{ inputs.package-name }}/README.md
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      shell: bash
+      run: |
+        ### Signed commit workaround - if we do a normal `git commit` here, it will be unsigned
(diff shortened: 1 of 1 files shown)


.github/actions/action.yml
```diff
@@ -87,26 +87,34 @@ runs:
       with:
         ref: ${{ github.event.pull_request.head.ref }}
         repository: ${{ github.event.pull_request.head.repo.full_name }}
-    - name: Check for terraform-docs.yml
```
amitchell-moz, line 90: Why is this part getting deleted? Without this, changes to any folder without a `.terraform-docs.yml` will fail CI

amitchell-moz: If you revert the deletion of this section and restore the `if: ${{ steps.tf_docs.outputs.tf_docs == 'true' }}`, then you don't need a `.terraform-docs.yml`.

The logic here is "if tf-docs.yml exists in the module folder, regenerate the readme with it; otherwise do nothing"

jbuck: So, if I want docs automatically generated for each PR, I should add `.terraform-docs.yml` right?

Follow-up question - should I make this PR *only* to get commit signing working (and thus, restore the `if` check)? And then I can use #256 to have documentation generated automatically for all modules?

amitchell-moz: So if you want all of the modules in https://github.com/mozilla/terraform-modules/pull/256/files to get auto-readmes, they would need `.teraform-docs.yml` in addition to those `BEGIN_TF_DOCS` blocks. 

I think you could just symlink in the `.teraform-docs.yml` from the root of the repo if you want to save a little work.

jbuck: Okay, reverted the `if` change and this is ready for re-review!

amitchell-moz (approved):
LGTM - thanks for fixing this!

PR #259, opened 2025-02-04
feat(google_gke): added configuration for control_plane_endpoints_config, updated versions and added variable to toggle

dlactin:
## Changelog entry
```
Adding control_plane_endpoints_config configuration, private access to the control plane is enabled by default, and a variable toggles public access. 

This will allow authorized users to access the control plane of our clusters without using our bastions.
```
Docs: https://cloud.google.com/blog/products/containers-kubernetes/new-dns-based-endpoint-for-the-gke-control-plane


jasonthomas (approved):

jasonthomas:
Need to fix the CI checks before merge.

jbuck (approved):

PR #260, opened 2025-02-04
Adding additional permissions for FxA devs to be able to view job his…

rachaelcrook:
…tory for datasets in BQ as well as cloudtask queues. This is specifically for FxA to be able to support the inactive account deletion script that is being implemented

## Changelog entry
```
Adding the following roles to other_roles.tf and outputs.tf
 roles/cloudtasks.queueAdmin
 roles/cloudtasks.taskRunner
roles/bigquery.resourceViewer
```

### DIFF
--- a/google_permissions/outputs.tf
+++ b/google_permissions/outputs.tf
@@ -5,7 +5,7 @@ locals {
   folder_additional_roles = [
     "roles/bigquery.jobUser",
     "roles/bigquery.dataViewer",
-    "roles/bigquery.resourceAdmin",
+    "roles/bigquery.resourceViewer",
     "roles/redis.admin",
     "roles/logging.admin",
     "roles/monitoring.alertPolicyEditor",
--- a/google_permissions/outputs.tf
+++ b/google_permissions/outputs.tf
@@ -5,6 +5,7 @@ locals {
   folder_additional_roles = [
     "roles/bigquery.jobUser",
     "roles/bigquery.dataViewer",
+    "roles/bigquery.resourceAdmin",
     "roles/redis.admin",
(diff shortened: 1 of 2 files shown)


rachaelcrook:
@chenba

google_permissions/outputs.tf
```diff
@@ -5,6 +5,7 @@ locals {
   folder_additional_roles = [
     "roles/bigquery.jobUser",
     "roles/bigquery.dataViewer",
+    "roles/bigquery.resourceAdmin",
```
jasonthomas, line 8:
```suggestion
    "roles/bigquery.resourceAdmin",
```

Is this the correct role we are looking for here? This role is associated with managing slot capacity, which doesn't happen in tenant projects.

If it's for managing and manipulating datasets and tables you might want 
https://cloud.google.com/bigquery/docs/access-control#bigquery.dataEditor

jasonthomas: I missed the truncated comment, you want to be able to view bigquery job details, roles/bigquery.resourceViewer might be better here.

rachaelcrook: I was going off of https://cloud.google.com/bigquery/docs/managing-jobs#required_permissions_2 which mentions roles/bigquery.resourceAdmin which has the bigquery.jobs.listAll capability. However, roles/bigquery.resourceViewer also has bigquery.jobs.listAll. So if that works then that sounds even better/more granular

rachaelcrook: Updated it to use resourceViewer.

jasonthomas (approved):

PR #261, opened 2025-02-04
moving cloudtask roles to folder additional roles instead of project …

rachaelcrook:
…to fix syntax issue

putting roles/cloudtasks.queueAdmin and roles/cloudtasks.taskRunner in the folder_additional_roles instead of project_additional_roles to allign with how it's being set in other_roles.tf.

## Changelog entry
```
Moving
 roles/cloudtasks.queueAdmin 
 roles/cloudtasks.taskRunner
to
 folder_additional_roles instead of project_additional_roles
```


rachaelcrook:
@chenba

cvalaas (approved):

PR #264, opened 2025-02-06
chore(google_permissions) - added roles needed for @mozilla.com mops users - OPST-1891

tcotav:
This pull request introduces several changes to the `google_permissions/pam_entitlement.tf` file to support a custom role for the Privileged Access Manager (PAM) CLI user and to manage IAM roles for developers across non-production and production environments.

## Changelog entry
```
* Added a new custom IAM role `privileged_access_manager_custom_role` with specific permissions tailored for the PAM CLI user, excluding delete entitlement permissions. This will be added in each project.
* Added IAM role assignments for `serviceusage.serviceUsageAdmin` to developers in both non-production and production projects to enable gcloud auth commands and grant creation for PAM entitlements.
* Added IAM role assignments for the newly created `privileged_access_manager_custom_role` to developers in both non-production and production projects.
```


jasonthomas (dismissed):

jasonthomas:
Although I am not exactly sure if we still need these additional roles is that right?

tcotav:
> Although I am not exactly sure if we still need these additional roles is that right?

@jasonthomas 

Coming back to this -- ends up we do need this role so that @mozilla.com users have the ability to use the console and revoke their grants or do anything with their entitlements. I think what we went back and forth on with this was where we were going to create a custom role. That ended up at the org level.

current bug: https://mozilla-hub.atlassian.net/browse/OPST-2182 - this I think is a symptom of that

I'm going to merge this.

Though, this assumes that it is the developer subgroup from a tenant workgroup. If (I should just say when) teams create custom subgroups, they won't have this permissions to manipulate their grants until this role is granted. Thoughts on that? I can open a bug for that specifically and close this now?

tcotav:
One more change incoming related to other changes we made in the interim. Also a big merge to pull all of that in.

jasonthomas:
> Though, this assumes that it is the developer subgroup from a tenant workgroup. If (I should just say when) teams create custom subgroups, they won't have this permissions to manipulate their grants until this role is granted. Thoughts on that? I can open a bug for that specifically and close this now?

I'd say let's open up a bug for that. I think it goes back to moving this permission management out of being a "tenant" concern into a "platform" concern, where we can grant this custom JIT role on subgroup creation, rather than the dance we do right now.

jasonthomas (approved):

PR #270, opened 2025-03-05
chore:run pre-commit

amitchell-moz:
## Changelog entry
```
Run pre-commit to fix trailing whitespace
```

Will also generate 0.0.3 tag which should have come with https://github.com/mozilla/terraform-modules/pull/267, now that https://github.com/mozilla/terraform-modules/pull/269 is merged


whd (approved):
https://www.gnu.org/software/emacs/manual/html_node/emacs/Useless-Whitespace.html

PR #280, opened 2025-03-28
feat(zizmor): Add Zizmor GitHub Action

bkochendorfer:
Zizmor is a Static Analysis tool for GitHub Actions https://woodruffw.github.io/zizmor/usage/#use-in-github-actions

## Changelog entry
```
TODO: Replace this inner text with a useful message
for users of the affected modules!
```

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -1,0 +1,29 @@
+# https://github.com/woodruffw/zizmor
+name: GitHub Actions Security Analysis with Zizmor
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["*"]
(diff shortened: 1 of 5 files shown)


.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Zizmor latest via Cargo
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Repository
+        uses: actions/checkout@v2
```
jbuck, line 22:
```suggestion
        uses: actions/checkout@v4
```

jbuck (approved):

PR #285, opened 2025-04-10
feat(google_workload_identity): removed data source and updated outputs to use provided service account name and email

dlactin:
## Changelog entry
```
* Removing data source for existing google_service_account, this is failing in hashicorp/google versions 6.13+ 
* Using `gcp_sa_name` directly instead of data source 
* Adding new `gcp_sa_email` variable that can be populated for existing service accounts
```
This update requires accompanying changes to the `google_gke_tenant` module to use the service account name instead of the email for existing service accounts.


dlactin:
Adding a major label as this might require additional changes to consume downstream.

bkochendorfer (approved):
This seems good to me. Is there a reason/ticket somewhere I can look at?

dlactin:
> This seems good to me. Is there a reason/ticket somewhere I can look at?

Our recent provider updates broke tenant bootstrapping. This data source underwent some changes in 6.13+, but we also have direct access to the service account name and email in the modules consuming it. It seemed reasonable to pass in both names and emails instead of trying to look up one component of the service account.

Ticket: MZCLD-3

PR #289, opened 2025-04-15
feat(google_monitoring): Add alert policies support for uptime checks.

mikaeld:
This module change is based off the work from this [PR](https://github.com/mozilla-it/dataservices-infra/pull/686). This adds Alert Policies associated to Uptime Checks. Supports custom documentation, links, etc. This is a major revision because new variables are required.

# Related PR
* Test PR: https://github.com/mozilla-it/dataservices-infra/pull/694

## Changelog entry
```
Add support for Alert Policies associated with Uptime Checks
```


quiiver:
@mikaeld I think it would be useful if we had a default check and alert policy so that all that was needed would be to drop the module into the tenant tf with very minimal config. i believe this to be useful for a couple of reasons. We should aim to reduce decision overhead for application owners. additionally, we have the opportunity to define an initial configuration that fits within what we would consider best practices.

mikaeld:
@quiiver I added an example with minimal configuration with just a few lines of config. For now, the default configuration I've set in this module matches the expected SLO for the services I support, and it ensures we stay within the free tier for uptime checks. For now, if you want to add checks for a service that's client facing/business critical, you need to override the default configuration as this only alerts after 5 minutes of down time. I think the obs team can weigh in on sane defaults, either in this PR or in a subsequent one.

To make it even more dev-friendly, @whd suggested the possibility to have the configuration (or configuration tier) defined in a tenant YAML file.

google_monitoring/README.md
```diff
@@ -1,12 +1,147 @@
...
+  labels = {
+    email_address = "dev_team+alerts@mydomain.com"
+  }
+  force_delete = false
+}
+
+module "uptime_checks" {
+  # Using main branch for simplicity, always pin your dependencies
+  source     = "github.com/mozilla/terraform-modules//google_monitoring?ref=main"
+  project_id = local.project_id
```
whd, line 17: Nit: this isn't `terraform fmt`'d

google_monitoring/README.md
```diff
@@ -1,12 +1,147 @@
...
+  environment  = local.environment
+  application  = local.application
+  realm        = local.realm
+
+  uptime_checks = [
+    {
+      name = local.uptime_check_name
+      host = myservice.mydomain.com
```
whd, line 25:
```suggestion
      host = "myservice.mydomain.com"
```

google_monitoring/README.md
```diff
@@ -1,12 +1,147 @@
...
```
whd, line 107:
```suggestion
      host = "myservice.mydomain.com"
```

whd (commented):
The README examples look good to me and definitely provide more structure than anything I'm aware of currently, but I could see it being a bit much for random users to grok. It would be nice to make it even simpler e.g. if it's called in a context where dev teams and service endpoints are already available so they don't need to be hard-coded. Or as we talked about some kind of YAML layer etc.

But I don't think we need to be too careful about designing this interface at least for the DSRE transition stuff because we're basically just dealing with simple cases.

Note: I didn't check for backwards compatibility with this module but it should ideally not do anything unless explicitly configured to.

This is all drive-by commentary: I don't think I'm the right person for primary review as this feels like it's more in the Obs domain than my typical purview.

bkochendorfer (approved):

PR #291, opened 2025-04-22
feat(google_cloudsql_mysql): add support for ENTERPRISE_PLUS on replicas

emaydeck-mozilla:
## Changelog entry
```
This allows users to enable Cloud SQL Enterprise Plus on replica instances. To configure this, set `replica_edition = "ENTERPRISE_PLUS"` in upstream modules.

I also alphabetized variables.tf to make it easier to find things and explicitly defined types for each variable to get rid of warnings in IDEs.
```

### DIFF
--- a/google_cloudsql_mysql/README.md
+++ b/google_cloudsql_mysql/README.md
@@ -110,8 +110,8 @@ output "mysql_database" {
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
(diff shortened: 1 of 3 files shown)


jbuck (approved):
Could you add `replica_edition` to the Postgres module as well?

Also I wish we had a single database module...

emaydeck-mozilla:
@jbuck I'm going to put that in a separate PR since I alphabetized variables.tf there too. It's beefy enough on its to be annoying if I include it here.

PR #303, opened 2025-05-06
fix(google_permissions): fix mozilla_workgroup module source

ahoneiser:
## Changelog entry
```
* fix source paths for `mozilla_workgroup` sub-modules in `google_permissions` module
```

### DIFF
--- a/google_permissions/workgroups.tf
+++ b/google_permissions/workgroups.tf
@@ -1,14 +1,14 @@
 module "admins_workgroup" {
-  source = "../mozilla_workgroup"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
   ids    = var.admin_ids
 }
 
 module "developers_workgroup" {
-  source = "../mozilla_workgroup"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
   ids    = var.developer_ids
 }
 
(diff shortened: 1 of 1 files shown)


google_permissions/workgroups.tf
```diff
@@ -1,14 +1,14 @@
 module "admins_workgroup" {
-  source = "../mozilla_workgroup"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
```
bkochendorfer, line 2: Is there a release of this we want to pin to?

ahoneiser: thought about that, too .. v2.6.1 is the most recent one .. I was unsure about adding this "subdependency" here, when we push other modules forward, this is likely going to remain at v2.6.1 in such a case .. :thinking: other modules we define point at the `main` ref, too, e.g. the tenant bootstrapping module does this .. :shrug:

ahoneiser: I bet this is one of those places that can be forgotten when the main module moves forward .. what would you recommend @bkochendorfer ?

ahoneiser: e.g. if I merge this, and create a new release of the repo, the pinned version would already be outdated ..

bkochendorfer (approved):

PR #307, opened 2025-05-23
feat:add override vars to bq log sink module

amitchell-moz:
## Changelog entry
```
Adds various overrides to bq log sink module to support v1 migrated infra
```
For https://mozilla-hub.atlassian.net/browse/SVCSE-2708

### DIFF
--- a/google_bigquery_log_sink/main.tf
+++ b/google_bigquery_log_sink/main.tf
@@ -3,22 +3,22 @@ locals {
 }
 
 resource "google_bigquery_dataset" "self" {
-  dataset_id = "log_storage_${var.environment}"
+  dataset_id = var.dataset_id_override != "" ? var.dataset_id_override : "log_storage_${var.environment}"
 
   default_table_expiration_ms     = local.ninety_days_in_ms
-  default_partition_expiration_ms = local.ninety_days_in_ms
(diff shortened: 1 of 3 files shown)


google_bigquery_log_sink/main.tf
```diff
@@ -3,22 +3,22 @@ locals {
 }
 
 resource "google_bigquery_dataset" "self" {
-  dataset_id = "log_storage_${var.environment}"
+  dataset_id = var.dataset_id_override != "" ? var.dataset_id_override : "log_storage_${var.environment}"
 
   default_table_expiration_ms     = local.ninety_days_in_ms
-  default_partition_expiration_ms = local.ninety_days_in_ms
+  default_partition_expiration_ms = var.use_partitioned_tables == true ? local.ninety_days_in_ms : 0
```
whd, line 9:
```suggestion
  default_partition_expiration_ms = var.use_partitioned_tables ? local.ninety_days_in_ms : 0
```

whd (approved):
LGTM. I would prefer using null values for defaults but I see we are pretty inconsistent about this in our modules.

`override` in the variable name kinda implies you shouldn't be setting this unless you have a good reason which is nice.

PR #311, opened 2025-06-05
feat(fastly): Dynamic condition and response_object

bkochendorfer:
## Changelog entry
```
The combination of a condition and response_object will allow you to serve a static maintenance page, or custom error responses, through Fastly.
```

### DIFF
--- a/google_fastly_waf/.terraform-docs.yaml
+++ b/google_fastly_waf/.terraform-docs.yaml
@@ -5,6 +5,7 @@ content: |-
 
   ```hcl
   {{ include "examples/example.tf" }}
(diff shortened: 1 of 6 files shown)


jbuck (approved):
Excellent, thank you for adding support for this! The only change I'd like to see is adding an example of the maintenance page in use to this module. Could you rip the code from https://github.com/mozilla-it/sandbox-infra/pull/637/files#diff-85776ccc69dab166576e1dd10bb797bf02d8541d1637387115704a8e99339b52 for that?

jbuck (approved):

PR #322, opened 2025-07-21
Adding read endpoint and read endpoint port to outputs

rachaelcrook:
## Changelog entry
```
Adding the read_endpoint and read_endpoint_port output variables. This is to be able to output read replica host and ports in tf configurations.
for users of the affected modules!
```


rachaelcrook:
> LGTM, thanks!
> 
> While you're here, could you also add the replica variables that we've used for Merino [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/merino/tf/prod/envs/prodpy/resources/redis.tf#L28-L29).

This will be done in the merino webservices repo under the redis config.

PR #341, opened 2025-08-13
feat(google_gke): add support for cluster-level high throughput logging

whd:
## Description

See https://cloud.google.com/kubernetes-engine/docs/how-to/adjust-log-throughput and https://mozilla-hub.atlassian.net/browse/OPST-1870 / https://mozilla-hub.atlassian.net/browse/SVCSE-3521 for motivation.

This is probably the simplest way to set up high throughput logging but it may not be flexible enough for our uses cases or current node pool configurations, as docs say:

>   Use a machine type with enough CPU to accommodate the high-throughput agent, such as e2-standard-8.

Also, per the same docs:
> Warning: Changing the logging agent on an existing node pool recreates all nodes. If you change the logging agent for an existing cluster, the existing node pools are unaffected and the change only applies to new node pools added to the cluster.

~~Setting this at the cluster level wouldn't affect any existing node pools anyway. If we want to affect the change we may need to explicitly set the value for each node pool.~~ EDIT: per :jbuck recommendation we set the global default and the node-specific value, all or nothing. This will cause all the node pools to roll all their nodes when it gets deployed, which is probably desirable but also something I will double check with folks about before proceeding.

If somebody with more experience managing this fairly substantial module wants to take over the PR or implementation please feel free to do so.

* OPST-1870


google_gke/cluster.tf
```diff
@@ -219,14 +219,13 @@ resource "google_container_cluster" "primary" {
...
+    for_each = var.enable_gcfs || var.enable_high_throughput_logging ? [1] : []
     content {
-      # TODO: If we end up needing to configure more parts of `node_pool_defaults`
-      # we will need to make this more dynamic
       node_config_defaults {
         gcfs_config {
-          enabled = true
+          enabled = var.enable_gcfs
         }
+        logging_variant = var.enable_high_throughput_logging ? "MAX_THROUGHPUT" : "DEFAULT"
```
jbuck, line 228: We could also set this in `google_container_node_pool.pools.node_config.logging_variant` to change it for all of the node pools as well

whd:
This should be a no-op because the default value remains the same. Per https://mozilla-hub.atlassian.net/browse/MZCLD-851?focusedCommentId=1123197 we're going to enable high-throughput logging in us-central1 as a basic test and then roll this out more broadly afterward, and maybe even change the default value once the rollout is complete.

PR #351, opened 2025-08-22
fix(fastly): Add better labels for GCP log components

bkochendorfer:
## Description

Trying to set some better labels for GCP logging artifacts created from module. Eventually this can be merged with the labels from the provider.

### DIFF
--- a/google_fastly_waf/bigquery.tf
+++ b/google_fastly_waf/bigquery.tf
@@ -24,7 +24,7 @@ resource "google_bigquery_dataset" "fastly" {
   }
 
   labels = {
-    env            = var.environment
+    env_code            = var.environment
     realm          = var.realm
     application    = var.application
     component_code = "fastly-logs"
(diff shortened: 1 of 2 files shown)


google_fastly_waf/bigquery.tf
```diff
@@ -22,6 +22,13 @@ resource "google_bigquery_dataset" "fastly" {
     role          = "WRITER"
     user_by_email = google_service_account.log_uploader.email
   }
+
+  labels = {
+    env            = var.environment
```
jasonthomas, line 27:
```suggestion
    env_code            = var.environment
```

google_fastly_waf/bigquery.tf
```diff
@@ -22,6 +22,13 @@ resource "google_bigquery_dataset" "fastly" {
...
+    realm          = var.realm
+    application    = var.application
```
jasonthomas, line 29:
```suggestion
    app_code    = var.application
```

PR #354, opened 2025-08-27
feat(fastly): Log Sampling

bkochendorfer:
## Description

Add the ability to enable log sampling for a service

### DIFF
--- a/google_fastly_waf/main.tf
+++ b/google_fastly_waf/main.tf
@@ -175,7 +175,7 @@ resource "fastly_service_vcl" "default" {
   }
 
   dynamic "condition" {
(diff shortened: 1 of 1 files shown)


google_fastly_waf/variables.tf
```diff
@@ -78,6 +78,16 @@ variable "service_account" {
   default = null
 }
 
+variable "log_sampling_percent" {
+  type    = string
+  default = "10"
```
whd, line 83: Does this count as a breaking change since it's not `100`? Also semantically it feels like this should be a number but I get that it's being interpolated into a string.

whd (approved):
I assume the new default is desirable

PR #356, opened 2025-08-27
feat(google_provider): Create Google provider wrapper that includes the default labels

tcotav:
## Description

This PR adds a wrapper for both the google_provider and the google_provider_beta terraform providers.

This is to add our expected values to the default_labels variable supported by the providers. Additionally, this will allow us to centralize which version of the providers is currently supported.

* MZCLD-759
* MZCLD-906

### DIFF
--- a/google_provider/README.md
+++ b/google_provider/README.md
@@ -2,6 +2,7 @@
 
 | Name | Version |
 |------|---------|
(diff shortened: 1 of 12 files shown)


tcotav:
the tests confirming this were faulty - going to delete it

PR #379, opened 2025-09-22
fix(google_fastly_waf): Allowing force domains update on TLS subscription

jbuck:
## Description
While working on the Bugzilla incident I had to update the domain list. Terraform wouldn't let me make changes to the certificate without setting this parameter, so I had to do it through the web console instead. This PR allows folks to make potentially unsafe domain changes through Terraform.

* INFRASEC-1228


bkochendorfer (approved):
cool thank you!

PR #380, opened 2025-09-24
fix(monitor): Allow redis instance to be destroyed

jwhitlock:
Allow setting `lifecycle.prevent_destroy` to false, so a redis instance can be destroyed.
~Also, fix typos and other issues found by `pre-commit`.~ - moved to PR #381


jwhitlock:
OK, `terraform` does *not* allow variables in the [lifecycle block](https://developer.hashicorp.com/terraform/language/block/resource#lifecycle), since it gets processed differently. You can use an [override file](https://developer.hashicorp.com/terraform/language/files/override) for local terraform but [not for modules](https://mozilla.slack.com/archives/C08CLU5LJK0/p1758733320734709?thread_ts=1758725286.744689&cid=C08CLU5LJK0).

We'll try some combo of a [removed block](https://developer.hashicorp.com/terraform/language/block/removed) and manual deletion instead

PR #455, opened 2026-04-09
fix(waf): Update to latest Fastly and change bot_management

bkochendorfer:
## Description

Behavior was changed over in https://github.com/fastly/terraform-provider-fastly/pull/1221

* MZCLD-2653

### DIFF
--- a/google_fastly_waf/versions.tf
+++ b/google_fastly_waf/versions.tf
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     fastly = {
       source  = "fastly/fastly"
-      version = ">= 5.13.0, < 9.0.0"
+      version = ">= 9.0.0"
     }
     sigsci = {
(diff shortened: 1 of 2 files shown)


bkochendorfer:
Tested in https://github.com/mozilla/sandbox-infra/pull/927

google_fastly_waf/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     fastly = {
       source  = "fastly/fastly"
-      version = ">= 5.13.0, < 9.0.0"
+      version = ">= 9.0.0"
```
amitchell-moz: Can you use "~> 9.0" instead here? That should prevent breakages like this going forward

quiiver (dismissed):

jbuck (dismissed):

jbuck (approved):

PR #465, opened 2026-05-06
feat(google_fastly_waf): Allow migrating away from EdgeDeployment APIs

Tibap:
## Description

Adds a `legacy_edge_deployment` variable to allow choosing between the legacy SigSci EdgeDeployment APIs and the newer Fastly APIs. This enables existing services to adopt the latest version without requiring an immediate migration away from the legacy backend.

With new Fastly's APIs:
- A new `product_enablement` is used to enabled ngwaf
- `EdgeDeployment` resources are no longer supported
- Setting `req.http.X-Sigsci-Response-Headers = "true"` are not needed anymore
- `Edge Security` dictionary is replaced by the new product `enablement` section
- SigSci headers are replaced in bigquery logging

* [INFRASEC-2932](https://mozilla-hub.atlassian.net/browse/INFRASEC-2932)


bkochendorfer (approved):
Looks good to me.

PR #469, opened 2026-05-14
feat(tf): implement new versioning scheme; update readme

amitchell-moz:
## Description
- Converts all pessimistic pins into equivalent bracketed pins to match new versioning strategy (https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/2606923830/Tofu+Terraform+Versioning+Strategy).
- Updates readme to reflect changes
- Bump to latest composite action version to pull in `google_project` fix (https://github.com/mozilla/terraform-modules/commit/22e5e067f6867f361158c21ad3034fa2aa28576c)

* MZCLD-696

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This repository also uses [GitHub Actions](.github/workflows/ci.yaml) to run sta
   * the module subdirectory name follows the format `provider_name`;
   * a `README.md` exists (use terraform-docs, detailed below, to generate);
   * a `main.tf` file exists;
-  * a `versions.tf` file exists & contains value `terraform.required_version`;
+  * a `versions.tf` file exists
 * security checks are run, including validating providers used are allowed & secrets scanning (via same pre-commit tooling).
 
 ### Versioning and Releases
@@ -27,6 +27,10 @@ If your PR contains [Conventional commits](https://www.conventionalcommits.org/e
 
 Release changelogs are generated based off of the contents of your PR description.
 Readmes will be automatically generated, using `.terraform-docs.yml` from the module directory (if it exists).
+
+#### Provider Versions
+
+See https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/2606923830/Tofu+Terraform+Versioning+Strategy for guidance on provider versioning.
 
 ## Creating modules
 
@@ -67,16 +71,17 @@ Some examples on using modules in this repository follow.
 }
 ```
 
-Using the (imaginary) `google_gke-cluster` module with always the latest version (e.g. following main branch):
-```terraform
- module "gke" {
-  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=main"
-}
-```
-
 Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary) git commit:
 ```terraform
  module "gke" {
   source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=69ad17030bfa4ea46f68f8cc449102d446658851"
+}
+```
+
+**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI
+Using the (imaginary) `google_gke-cluster` module with always the latest version (** Not recommended**):
+```terraform
+ module "gke" {
+  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=main"
 }
 ```
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary
 }
 ```
 
-**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI
+**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI.
 Using the (imaginary) `google_gke-cluster` module with always the latest version (**Not recommended**):
 ```terraform
  module "gke" {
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -67,16 +71,17 @@ Some examples on using modules in this repository follow.
...
 ```terraform
  module "gke" {
-  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=main"
+  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=69ad17030bfa4ea46f68f8cc449102d446658851"
 }
 ```
 
-Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary) git commit:
+**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI
+Using the (imaginary) `google_gke-cluster` module with always the latest version (** Not recommended**):
```
whd:
```suggestion
Using the (imaginary) `google_gke-cluster` module with always the latest version (**Not recommended**):
```

README.md
```diff
@@ -67,16 +71,17 @@ Some examples on using modules in this repository follow.
...
+Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary) git commit:
 ```terraform
  module "gke" {
-  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=main"
+  source      = "git@github.com:mozilla/terraform-modules.git//google_gke-cluster?ref=69ad17030bfa4ea46f68f8cc449102d446658851"
 }
 ```
 
-Using the (imaginary) `google_gke-cluster` module based on a specific (imaginary) git commit:
+**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI
```
whd: or add blank line, not sure how you want to format this

```suggestion
**Do not pin to main**: pinning to main makes your infrastructure non-deterministic, this will soon be disallowed by CI.
```

whd:
The `~>` to bracketed-range conversions aren't symmetric across the diff, and a few look like they deviate from the intent of the [versioning strategy](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/2606923830/Tofu+Terraform+Versioning+Strategy).

### `~> X.0` conversions widen the range (probably intended)

For the AWS and Google modules, the conversion widens the allowed set:

- `aws_gcp_vpn/versions.tf` aws: `~> 6.0` previously allowed `6.0.0` through `6.99.99`. The new `> 5.0, < 7.0` now also allows `5.0.1`, `5.42.0`, etc. So a fresh `terraform init` against this module could resolve to aws provider `5.99.0` where before it would have resolved to `6.x`.
- Same pattern for google `~> 7.0` becoming `> 6.0, < 8.0`: `6.0.1` and `6.42.0` are now in range where before only `7.x` was.

This matches the "N-1 and N" rule, but it is a real behavioral change for any consumer not using a `.terraform.lock.hcl`.

### `~> 3.30` narrows the range (likely a bug)

`pagerduty_service/versions.tf` and `pagerduty_team/versions.tf`:

- Old `~> 3.30` allowed `3.30.0`, `3.31.0`, `3.42.7`, up to but not including `4.0.0`.
- New `> 3.29, < 3.31` allows `3.29.1`, `3.30.0`, `3.30.5`, but blocks `3.31.0` and everything above.

So a consumer currently on pagerduty `3.31.2` (which satisfied `~> 3.30`) will fail `terraform init` after this PR lands. Looking at the [pagerduty provider releases](https://github.com/PagerDuty/terraform-provider-pagerduty/releases), `3.31`, `3.32`, `3.33` and `3.34` all exist, so this is not hypothetical.

### `~> 0.12` narrows the range (likely the same bug)

`google_psc_to_elastic/versions.tf` ec provider:

- Old `~> 0.12` allowed `0.12.0`, `0.13.0`, `0.99.0`, up to but not including `1.0.0`.
- New `> 0.11, < 0.13` allows `0.11.1`, `0.12.0`, `0.12.5`, but blocks `0.13.0` and above.

### Root cause

Looks like a mechanical `~> A.B` to `> (A-1).B, < (A+1).B` substitution. That rule is only correct when `B == 0` (i.e. `~> 7.0`, where the second component is genuinely the next-tier boundary). For `~> 3.30` and `~> 0.12`, the second component is the minor version, not the major, so subtracting and adding 1 to it produces a single-minor pin instead of an N-1/N major range.

### `>` is strict, not `>=`

`> 5.0` in Terraform means strictly greater than `5.0.0`, so version `5.0.0` itself is excluded. Concrete impact: if aws ever ships exactly `5.0.0` and someone wants to use it, `aws_gke_oidc_config` would refuse. Probably not load-bearing in practice for the providers in this PR, but if the strategy is "support major N-1 and N inclusive," the lower bound should be `>=`:

```hcl
# instead of:
version = "> 5.0, < 7.0"
# write:
version = ">= 5.0, < 7.0"
```

### Suggested fix

Re-derive each constraint from the strategy doc rather than from the existing pin string, and prefer `>=` for the lower bound.

| File | Current PR | Suggested |
|---|---|---|
| `aws_*/versions.tf` (aws) | `> 5.0, < 7.0` | `>= 5.0, < 7.0` |
| `google_*/versions.tf` (google) | `> 6.0, < 8.0` | `>= 6.0, < 8.0` |
| `pagerduty_*/versions.tf` | `> 3.29, < 3.31` | `>= 3.0, < 4.0` (or whatever the strategy says) |
| `google_psc_to_elastic/versions.tf` (ec) | `> 0.11, < 0.13` | `>= 0.11, < 0.13` if "pin to 0.12-ish," otherwise widen per strategy |
| `mozilla_access_event_consumer/versions.tf` (archive) | `> 1.0, < 3.0` | `>= 1.0, < 3.0` |
| `google_certificate_manager_certificate_map/versions.tf` (random) | `> 2.0, < 4.0` | `>= 2.0, < 4.0` |

whd (dismissed):
r+ though see the 🤖 analysis. I'm not particularly worried about any temporary pain this causes as we are trying to get to a place where the ongoing pain is reduced.

whd (approved):

PR #472, opened 2026-05-18
feat(codeowners): add gcp-wg to all CODEOWNERs

amitchell-moz:
## Description
Adds `gcp-wg` to every CODEOWNER entry

* https://mozilla-hub.atlassian.net/browse/MZCLD-2700


whd (approved):
Some of these are pretty redundant now but we can refactor that separately

PR #475, opened 2026-05-28
fix(versions): rm last main pins; fix examples

amitchell-moz:
## Description
Removes last remaining `main` pins in `google_bigquery_syndicated_dataset` & `google_permissions` to comply with our [versioning strategy](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/2606923830/Tofu+Terraform+Versioning+Strategy).

Also replaces all main pins in examples.

* MZCLD-2019

### DIFF
--- a/google_bigquery_syndicated_dataset/main.tf
+++ b/google_bigquery_syndicated_dataset/main.tf
@@ -98,7 +98,7 @@ data "terraform_remote_state" "org" {
 # Service accounts that perform syndication
 # Currently Jenkins with plans to move to Airflow, see https://mozilla-hub.atlassian.net/browse/SVCSE-3005
 module "syndication_workgroup" {
-  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=mozilla_workgroup-1.3.0"
   ids    = var.syndication_workgroup_ids
 }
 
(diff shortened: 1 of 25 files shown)


google_bigquery_syndicated_dataset/main.tf
```diff
@@ -98,7 +98,7 @@ data "terraform_remote_state" "org" {
 # Service accounts that perform syndication
 # Currently Jenkins with plans to move to Airflow, see https://mozilla-hub.atlassian.net/browse/SVCSE-3005
 module "syndication_workgroup" {
-  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=mozilla_workgroup-1.3.0"
```
jbuck, line 101: should this reference the local path instead?

google_permissions/workgroups.tf
```diff
@@ -1,14 +1,14 @@
 module "admins_workgroup" {
-  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=main"
+  source = "github.com/mozilla/terraform-modules//mozilla_workgroup?ref=mozilla_workgroup-1.3.0"
```
jbuck, line 2: Should this reference the local path instead?

jbuck (approved):
r+ with 2 suggestions

google_bigquery_syndicated_dataset/main.tf
amitchell-moz: For modules that aren't tightly coupled (like the `aws_gke_oidc_*`s) I think it makes more sense to pin - that way breaking changes can be made to `mozilla_workgroup` without having to think about bigquery datasets

PR #480, opened 2026-06-09
feat(google_deployment_accounts): add gha_branch_assertions

whd:
## Description

This is a no-op against existing configuration but introduces some new GHA options for branch-based OIDC configuration. See the Jira ticket for details.

Most of this is copypasta from the similar CircleCI config. There's a slight difference in precedence for GHA since there's no CircleCI context to consider but I expect only a couple people to ever need to do any advanced configuration.

I reordered the examples to make it clearer that GHA is the preferred CI system.

I've also updated CODEOWNERS to remove the individual members of gcp-wg since they already get reviews on all the things, and in the spirit of https://mozilla-hub.atlassian.net/browse/MZCLD-3214 removed Sven's old CODEOWNER access.

https://github.com/mozilla/global-platform-admin/pull/6590 was required to add the new `repository_ref` attribute upon which this is based.

* MZCLD-924

### DIFF
--- a/google_deployment_accounts/main.tf
+++ b/google_deployment_accounts/main.tf
@@ -23,9 +23,11 @@ resource "google_service_account_iam_binding" "github-actions-access" {
 
 locals {
   circleci = var.wip_name == "circleci"
-  # explicit attributes replace all other kinds of assertions
+  # explicit attributes replace all other kinds of assertions. A subject
+  # specifier maps to a single identity (principal://); all other attributes
+  # map to identity sets (principalSet://).
   circleci_attribute_assertions = local.circleci ? [for attribute_specifier in var.circleci_attribute_specifiers :
-    "principalSet://iam.googleapis.com/projects/${var.wip_project_number}/locations/global/workloadIdentityPools/${var.wip_name}/${attribute_specifier}"
+    "${split("/", attribute_specifier)[0] == "subject" ? "principal" : "principalSet"}://iam.googleapis.com/projects/${var.wip_project_number}/locations/global/workloadIdentityPools/${var.wip_name}/${attribute_specifier}"
   ] : []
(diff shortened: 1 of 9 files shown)


google_deployment_accounts/variables.tf
```diff
@@ -26,6 +26,36 @@ variable "gha_environments" {
...
+  default     = []
+}
+
+variable "gha_attribute_specifiers" {
+  description = "Set of attribute specifiers to allow deploys from, in the form ATTR/ATTR_VALUE. If specified, this overrides the github_repository variable and any other GHA-specific variables."
+  type        = set(string)
+  default     = []
+  validation {
+    condition = alltrue(
+      [for attribute_specifier in var.gha_attribute_specifiers :
```
whd, line 45: This set of attributes is statically encoded for additional safety but it does introduce an extra manual step if we add and rely on new attributes. I believe the additional plan-time safety is warranted especially since attribute updates are uncommon.

google_deployment_accounts/main.tf
```diff
@@ -23,9 +23,11 @@ resource "google_service_account_iam_binding" "github-actions-access" {
 
 locals {
   circleci = var.wip_name == "circleci"
-  # explicit attributes replace all other kinds of assertions
+  # explicit attributes replace all other kinds of assertions. A subject
+  # specifier maps to a single identity (principal://); all other attributes
```
whd, line 27: The 🤖 caught this edge case that wasn't accounted for before. It doesn't affect any existing config and isn't likely to affect any config in the future, but it is technically correct to make this change so I'm fixing it for CircleCI and the new GHA logic has the same logic.

jasonthomas (approved):

jbuck (approved):

PR #486, opened 2026-08-05
feat(google_gke): add insecure_kubelet_readonly_port_enabled

bkochendorfer:
Adds `insecure_kubelet_readonly_port_enabled` (bool, default `false`) to the `google_gke` module, wiring it into both:

- `google_container_cluster.primary.node_pool_defaults.node_config_defaults` — the default for node pools created after this is applied
- `google_container_node_pool.pools.node_config.kubelet_config` — every pool the module manages

Both are needed: GKE does not retrofit the cluster-level default onto existing node pools.

The `node_pool_defaults` block is no longer conditional on `enable_gcfs`/`enable_high_throughput_logging`, since the read-only port setting always needs to be present. `gcfs_config` moved to a `dynamic` block so behavior is unchanged when `enable_gcfs = false`.

### Behavior change for consumers

The default is `false`, so bumping to this version disables port 10255 on all node pools. Consumers should stage the rollout per cluster/region rather than applying fleet-wide in one go, and can set `insecure_kubelet_readonly_port_enabled = true` to defer.

### Context

Unblocks [MZCLD-1982](https://mozilla-hub.atlassian.net/browse/MZCLD-1982), which has been on hold since Sept 2024 waiting for provider support. The provider fields landed in google/google-beta 6.x; the module already pins `>= 6.0, < 8.0`, so no provider bump is required.

Ref: https://cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port

### Testing

`tofu fmt -check` and `tofu validate` clean. No apply run yet — first apply will be against `sandbox-high-nonprod`.


jbuck (dismissed):

google_gke/cluster.tf
```diff
@@ -227,14 +227,17 @@ resource "google_container_cluster" "primary" {
...
 
-  dynamic "node_pool_defaults" {
-    for_each = var.enable_gcfs || var.enable_high_throughput_logging ? [1] : []
-    content {
-      node_config_defaults {
-        gcfs_config {
+  # Defaults for new node pools only; existing pools are set per pool below.
+  node_pool_defaults {
+    node_config_defaults {
+      insecure_kubelet_readonly_port_enabled = var.insecure_kubelet_readonly_port_enabled ? "TRUE" : "FALSE"
```
jasonthomas: I don't think we should make this configurable.

jbuck (approved):

PR #487, opened 2026-08-07
feat(fastly): Log response `Content-Encoding` into BigQuery

leplatrem:
## Description

Without this, we don't have any way to track which content-encoding ended up being negociated with the client.

In the context of Dictionary Compression Transport, we need to distinguish clients that pull deltas, from gzip, and from raw responses.

Alternative approach was this : https://github.com/mozilla/webservices-infra/pull/12059 but adding this column seems cleaner 

* https://mozilla-hub.atlassian.net/browse/RMST-486


alexcottner (approved):

Tibap:
Hi @leplatrem, thanks for this PR. 

Adding a column like this will impact the schema of all other Bigquery tables which can cost us a lot in the end. Before introducing this change, we need to rework how `bigquery.tf` builds the schema so that columns can be added conditionally. That’s a larger change and goes beyond the scope of this PR.

I’m going to create a separate ticket to track that work. In the meantime, I think we should put this PR on hold until that change is in place.

Tibap:
As a side note, once the larger change is implemented, this PR will also need to edit `bq_format_v2.txt` to include this new column.

Tibap:
@leplatrem I think we can close this PR now. See my comment here for more information: https://github.com/mozilla/webservices-infra/pull/12059#issuecomment-5440936223

PR #494, opened 2026-09-07
feat(google_cloudsql_postgres): allow configuring backup start time

stevejalim:
Adds an optional backup_start_time variable (HH:MM, UTC) that maps to the backup_configuration.start_time argument on the primary instance.

Defaults to null, matching prior behavior where Cloud SQL picks a random start time, so this is a no-op for existing callers.


bqbn (approved):

stevejalim:
Thanks for the r+ @bqbn -- do you know if this can just be merged or if it need any kind of extra process (atlantis, for instance)?

